PR #31, opened 2016-10-18
connect with remote cluster

ycui1984:
tools (like jconsole) cannot connect with remote cluster without my fix


ycui1984:
@amnonh 
this pull request solves connection issue with scylla.
Could you review and comment on it?

penberg:
@ycui1984 Thanks for the fix!

PR #32, opened 2016-10-18
update README to be consistent with code

ycui1984:


tzach:
Hi @ycui1984 
Thanks for the contribution!

scylla-jmx, like scylla, does not accept PR, just patch.
Can you please follow the instructions here:
https://github.com/scylladb/scylla#contributing-to-scylla
To submit you patch?

Thanks

PR #33, opened 2016-10-24
Scylla-jmx for C3

elcallio:
This is a pretty hefty change set. Base goal is to update JMX impl to Cv3 compatibility. While the base MBeans are easy enough, the metrics parts of the stew are trickier, since C3 changed a number of things around. 
To that end, and to clean up this code somewhat, this basically changes all about metrics management in scylla-jmx. Some background:

Cassandra uses codehaul metrics for actual _measurement_. Now, obviously, in this proxy project we need not measure anything, since actual happenings are in the scylla process. Previous version of the code however still utilized a (not-so-pretty reflection-hacked-into) version of codehaul metrics because they also provided the system for exposing the data through JMX. I.e. we added a bunch of stuff we really did not need, to avoid dealing with some of that we did. 

In C3, v3 of codehaul is used, which the Cassandra devs apparently did not like the JMX integration of. Thus they decided to deal with JMX exposure themselves. which makes sense, because they want to control the syntax/structure. But given this, we no longer have _any_ reason to utilize codehaul, since it does nothing for us. 

These change sets instead adapts the cassandra JMX bindings somewhat, and adds a wholly own structure of metric point binding, using java.util.function interfaces to provide flexible and late-ish binding to actual data query objects. End result is a much slimmer set of objects/functions to bind metrics (which of course are just queries to scylla api). 

Also, MX4J has been dropped, since it is at best broken. Instead, we use simple wrapping of the system management server object to deal with dynamically populating transients objects like column families. 

Removed most statefulness (beyond binding) in MBean impls, all "bookeeping" of sub-objects and bind status now uses the actual mbean server. I.e. remove race conditions + lighter bookkeep. 

Since this is Java, and everything is tied together in a ball of yarn, most of the changes here are _not_ self-contained. I.e. some of these will, applied individually, break the build. They are still kept as individual patches though, mainly for readability.


penberg:
@elcallio Can you please rebase the series to remove those version number changes that don't belong to this series? Then just force push to your branch and this pull request is cleaned up.

elcallio:
Ok, I've manually cleared away the offending commits.

avikivity:
Will this work with old nodetool and old scylla?

penberg:
@avikivity No. And it's tons of work to try to make that happen. Cassandra doesn't maintain backwards compatibility at the JMX level. I doubt it makes sense for us to try to do that either.

/cc @amnonh

avikivity:
Then how can we upgrade?

The upgrade procedure has old scylla and scylla-jmx running in parallel with new nodetool.

penberg:
Why? What is that required for?

slivne:
we may need to change the procedure to support this.

1. currently we install new version
2. nodetool drain scylla
3. restart jmx and scylla

we can do it differently

1. nodetool drain scylla
2. install new version
3. restart jmx and scylla

slivne:
running *"nodetool drain"*

elcallio:
On 2016-10-24 16:24, Avi Kivity wrote:
>
> Will this work with old nodetool and old scylla?
>
Old scylla, no. At least not fully. Some (most) calls/metrics will, but 
not all required by new node tool.
OId node tool should work mostly, as no calls were removed that were not 
already deprecated and unused in 2.x series.
But I'd like to point out that the previous decision was not _not_ 
attempt to keep 2.x series compatibility, since Cassandra does not 
(beyond basic compat for third party tools).

PR #45, opened 2017-08-22
fix auto_compaction request strings

glommer:
Column family is not being passed, so the requests fail to reach the
correct endpoint.

Signed-off-by: Glauber Costa <glauber@scylladb.com>

PR #46, opened 2017-08-22
fix auto_compaction request strings

glommer:
Column family is not being passed, so the requests fail to reach the
correct endpoint.

Signed-off-by: Glauber Costa <glauber@scylladb.com>


glommer:
The scylla counterpart won't be merged to 2.0, so we can drop this one. I still have a PR (#45) for the master branch for this

Issue #44, opened 2017-06-12
read parameters from configuration file


glommer:
Right now, if we want to  change the behavior of scylla-jmx, for example to accept remote connections, we need to change the systemd file itself.

That is both inconvenient and ugly by systemd standards. The best way to handle this is what we do in Scylla, using environment variables that are present in /etc/sysconfig/scylla-jmx.

The latter file already exists, but it is only passing some path variables. We should extend it to allow us to change scylla-jmx behavior (remote, ssl, auth, etc), through external configuration files.

penberg:
Fixed by commit https://github.com/scylladb/scylla-jmx/commit/e80a5e3cb319fc80c669d6d75f5abeb0d781758b from @fdv.

PR #48, opened 2017-09-26
Allow systemd to run scylla-jmx with params from /etc/defaults/

fdv:
Using systemd to run scylla-jmx won't load params from /etc/defaults/scylla-jmx  because it expects params to be sent using the command line.

This patch enhances the sysconfig file by adding the available options (commented) and passes the right options to systemd `ExecStart` when defined. That way scylla-jmx is ran with the params defined into /etc/defaults/scylla-jmx.

Maybe not the most elegant way to do it, but 1/ it works 2/ I didn't find a better solution to fix that problem.

### DIFF
--- a/dist/common/sysconfig/scylla-jmx
+++ b/dist/common/sysconfig/scylla-jmx
@@ -3,3 +3,29 @@ SCYLLA_HOME=/var/lib/scylla
 
 # scylla config dir
 SCYLLA_CONF=/etc/scylla
+
+SCYLLA_PARAMS=""
+
+# The jmx port to open
+# SCYLLA_PARAMS="${SCYLLA_PARAMS } -jp 7199"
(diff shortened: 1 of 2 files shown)


dorlaor:
@syuu1228 @amnonh pls review.
@fdv, thanks for the contribution! We usually use patches, not pulls but we can make it a one off, we do ask for a CLA - http://www.scylladb.com/open-source/

fdv:
I've sent the CLA, I can also send a patch the usual way if you prefer.

And this PR incidentaly fixes issues #44

penberg:
The PR is merged as the following commit:

https://github.com/scylladb/scylla-jmx/commit/e80a5e3cb319fc80c669d6d75f5abeb0d781758b

PR #67, opened 2019-06-23
ColumnFamilyStore: finishLocalSampling should respect count limit

### LINKED ISSUE #66: nodetool toppartitions should have a way for limiting the size
See https://github.com/scylladb/scylla/issues/4529

amnonh:
When calling nodetool toppartitions with size limit, finishLocalSampling
should respect that and limit the number of the results.

Example:
$ nodetool toppartitions -k 2 keyspace1 standard1 20
WRITES Sampler:
  Cardinality: ~2 (256 capacity)
  Top 2 partitions:
	Partition                Count       +/-
	38333032394d4f4d5030         4         3
	4e353937383137503330         4         3

READS Sampler:
  Cardinality: ~2 (256 capacity)
  Top 2 partitions:
	Nothing recorded during sampling period...

Fixes #66

### DIFF
--- a/src/main/java/org/apache/cassandra/db/ColumnFamilyStore.java
+++ b/src/main/java/org/apache/cassandra/db/ColumnFamilyStore.java
@@ -519,8 +519,8 @@ public class ColumnFamilyStore extends MetricsMBean implements ColumnFamilyStore
         JsonArray counters = tableSamplerResult.getJsonArray((samplerType.equalsIgnoreCase("reads")) ? "read" : "write");
         long size = 0;
         if (counters != null) {
-            size = counters.size();
-            for (int i = 0; i < counters.size(); i++) {
+            size = (count > counters.size()) ? counters.size() : count;
+            for (int i = 0; i < size; i++) {
                 JsonObject counter = counters.getJsonObject(i);
                 result.put(new CompositeDataSupport(COUNTER_COMPOSITE_TYPE, COUNTER_NAMES,
                         new Object[] { counter.getString("partition"), // raw


src/main/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -517,10 +517,12 @@ public CompositeData finishLocalSampling(String samplerType, int count) throws O
         TabularDataSupport result = new TabularDataSupport(COUNTER_TYPE);
 
         JsonArray counters = tableSamplerResult.getJsonArray((samplerType.equalsIgnoreCase("reads")) ? "read" : "write");
-        long size = 0;
+        long size = count;
```
nyh: If count==7 but counters.size() is empty, this will leave size=7 to be returned at the end. Is this fine or not?

amnonh: good catch, I'm sending a change that would cover this one

amnonh:
@nyh I forced push a new patch

src/main/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -519,8 +519,8 @@ public CompositeData finishLocalSampling(String samplerType, int count) throws O
         JsonArray counters = tableSamplerResult.getJsonArray((samplerType.equalsIgnoreCase("reads")) ? "read" : "write");
         long size = 0;
         if (counters != null) {
-            size = counters.size();
-            for (int i = 0; i < counters.size(); i++) {
+            size = (count > counters.size()) ? counters.size() : count;
```
nyh: std::min() would have been clearer, but I'm fine with this version too. I'll commit.

nyh:
Committed manually to next. Github doesn't know because of our wish not to do a merge commit in a single-patch series.

PR #69, opened 2019-06-30
ColumnFamilyStore: Mbean API support the hex format param

amnonh:
Cassandra 3.0 version of the JMX added a parameter that allows accepting
the parameter as hex.

This breaks the current implementation with a NoSuchMethodException.

This patch adds the missing implementation.

For a full support, a follow up patch in Scylla is needed, but for the
current functionality it would work.

After this patch usage example:

nodetool getsstables keyspace1 standard1 39303138374b4d343830

### DIFF
--- a/src/main/java/org/apache/cassandra/db/ColumnFamilyStore.java
+++ b/src/main/java/org/apache/cassandra/db/ColumnFamilyStore.java
@@ -369,6 +369,23 @@ public class ColumnFamilyStore extends MetricsMBean implements ColumnFamilyStore
     }
 
     /**
+     * Returns a list of filenames that contain the given key on this node
+     * @param key
+     * @param hexFormat if key is in hex string format
+     * @return list of filenames containing the key
+     */
+    @Override
+    public List<String> getSSTablesForKey(String key, boolean hexFormat)
+    {
+        log(" getSSTablesForKey(String key)");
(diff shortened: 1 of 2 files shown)


amnonh:
Fixes https://github.com/scylladb/scylla/issues/4619

amnonh:
ping

slivne:
@nyh ping

slivne:
@penberg nadav will be out for 2 weeks moving this to your queue

src/main/java/org/apache/cassandra/db/ColumnFamilyStore.java
```diff
@@ -368,6 +368,23 @@ public long estimateKeys() {
...
+     */
+    @Override
+    public List<String> getSSTablesForKey(String key, boolean hexFormat)
+    {
+        log(" getSSTablesForKey(String key)");
+        MultivaluedMap<String, String> queryParams = new MultivaluedHashMap<String, String>();
+        queryParams.add("key", key);
+        if (hexFormat) {
+            queryParams.add("hf", "True");
+        }
```
avikivity: hf? can we have a more descriptive name?

Like "format=text" or "format=hex"

amnonh: @avikivity sure, just forced push a new version

penberg:
@amnonh, I fixed merge target branch to `next`.

PR #70, opened 2019-07-01
make scylla-apiclient a separate module so the jar can be reused

tarzanek:
This works out of box
the only big difference is that now location of binary is
scylla-jmx/target 
(where previously it was just target/ )
I don't know how many projects us this as a submodule, so if they consume the binary directly I will need to fix them, too


elcallio:
This is an unnecessarily large change. You can just create the parent as a child-folder, and point to it with a relative path to avoid moving all source files (and impairing trackability).

penberg:
@tarzanek Please check with @hagitsegev and @syuu1228 on the impact of the move to `scylla-jmx/target` on packaging.

tarzanek:
@elcallio Done, this basically solves the concern with relocation, so this change should be transparent to everybody now, thank you Calle!

syuu1228:
@penberg @tarzanek To build .rpm/.deb packages with new target directory, we need to modify scripts/create-relocatable-package.py to copy the .jar into relocatable package.
Just change this line should be enough:
https://github.com/scylladb/scylla-jmx/blob/master/scripts/create-relocatable-package.py#L46

tarzanek:
@syuu1228 it should be OK after I changed the code based on what Calle suggested, so from my side this patch is done and working and not breaking anything

syuu1228:
@tarzanek Agree, and I can send a patch to change the path after this change get merged.

pom.xml
```diff
@@ -61,29 +48,25 @@
...
+<!--                        <manifest>-->
+<!--                            <mainClass>com.scylladb.jmx.main.Main</mainClass>-->
+<!--                            <addClasspath>true</addClasspath>-->
+<!--                            <classpathPrefix>lib/</classpathPrefix>-->
+<!--                        </manifest>-->
+<!--                        <manifestEntries>-->
+<!--                        </manifestEntries>-->
+<!--                    </archive>-->
+<!--                </configuration>-->
+<!--            </plugin>-->
```
penberg: Why is this commented out?

pom.xml
```diff
@@ -95,6 +78,8 @@
                             <goal>shade</goal>
                         </goals>
                         <configuration>
+<!--                            <shadedClassifierName>shaded</shadedClassifierName>-->
+<!--                            <shadedArtifactAttached>false</shadedArtifactAttached>-->
```
penberg: Ditto.

penberg:
There's some problem with this that seems to break package build:

```
[penberg@nero scylla-jmx]$ ./reloc/build_reloc.sh 
Last metadata expiration check: 0:13:56 ago on Mon 08 Jul 2019 10:59:20 AM EEST.
Package maven-1:3.5.4-5.module_f28+3939+dc18cd75.noarch is already installed.
Package java-1.8.0-openjdk-devel-1:1.8.0.212.b04-0.fc30.x86_64 is already installed.
Dependencies resolved.
Nothing to do.
Complete!
[INFO] Scanning for projects...
[INFO] 
[INFO] --------------------< com.scylladb.jmx:scylla-jmx >---------------------
[INFO] Building Scylla JMX 1.0
[INFO] --------------------------------[ jar ]---------------------------------
[WARNING] The POM for com.scylladb.jmx:scylla-apiclient:jar:1.0 is missing, no dependency information available
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 0.200 s
[INFO] Finished at: 2019-07-08T11:13:19+03:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal on project scylla-jmx: Could not resolve dependencies for project com.scylladb.jmx:scylla-jmx:jar:1.0: Failure to find com.scylladb.jmx:scylla-apiclient:jar:1.0 in https://repo.maven.apache.org/maven2 was cached in the local repository, resolution will not be reattempted until the update interval of central has elapsed or updates are forced -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/DependencyResolutionException

```

tarzanek:
ok, it seems maven submodule won't force the parent to build the other submodule
I will try to see if there is a way around

the other option would be to change ./reloc/build_reloc.sh  to run the install in the parent project
```
cd scylla-jmx-parent
mvn -B install
cd ..
```
instead of just `mvn -B install`

penberg:
Sure, we can fix up `build_reloc.sh` (and the instructions in README.md) I would prefer using the `--file` command line option over `cd`:

```
mvn -B --file scylla-jmx-parent/pom.xml install
```

PR #75, opened 2019-07-11
CompactionMetrics: use the pending compaction API

### LINKED ISSUE #74: get pending task by table should  use the API for get pending task

amnonh:
The PendingTasksByTableName metric should use the pending_tasks_by_table
API to get the real value of the pending compaction.

Fixes #74

### DIFF
--- a/src/main/java/org/apache/cassandra/metrics/CompactionMetrics.java
+++ b/src/main/java/org/apache/cassandra/metrics/CompactionMetrics.java
@@ -55,7 +55,7 @@ public class CompactionMetrics implements Metrics {
 
         registry.register(() -> registry.gauge((client) -> {
             Map<String, Map<String, Integer>> result = new HashMap<>();
-            JsonArray compactions = client.getJsonArray("compaction_manager/compactions");
+            JsonArray compactions = client.getJsonArray("compaction_manager/metrics/pending_tasks_by_table");
 
             for (int i = 0; i < compactions.size(); i++) {
                 JsonObject c = compactions.getJsonObject(i);
@@ -68,7 +68,7 @@ public class CompactionMetrics implements Metrics {
                 }
 
                 Map<String, Integer> map = result.get(ks);
-                map.put(cf, (int)(c.getJsonNumber("total").longValue() - c.getJsonNumber("completed").longValue()));
+                map.put(cf, (int)(c.getJsonNumber("task").longValue()));
             }
             return result;
         }), factory.createMetricName("PendingTasksByTableName"));


slivne:
@elcallio please review and give your blessing

slivne:
@penberg  - We have Calle's blessing

penberg:
I merged this, but @amnonh, can you explain where the `task` name for this bit:

```java
map.put(cf, (int)(c.getJsonNumber("task").longValue()));
```

Looking at the code in Scylla server's `api/compaction_manager.cc`, the API request is handled like this:

```c++
    cm::get_pending_tasks_by_table.set(r, [&ctx] (std::unique_ptr<request> req) {
        return ctx.db.map_reduce0([&ctx](database& db) {
            std::unordered_map<std::pair<sstring, sstring>, uint64_t, utils::tuple_hash> tasks;
            return do_for_each(db.get_column_families(), [&tasks](const std::pair<utils::UUID, seastar::lw_shared_ptr<table>>& i) {
                table& cf = *i.second.get();
                tasks[std::make_pair(cf.schema()->ks_name(), cf.schema()->cf_name())] = cf.get_compaction_strategy().estimated_pending_compactions(cf);
                return make_ready_future<>();
            }).then([&tasks] {
                return tasks;
            });
        }, std::unordered_map<std::pair<sstring, sstring>, uint64_t, utils::tuple_hash>(), sum_pending_tasks).then(
                [](const std::unordered_map<std::pair<sstring, sstring>, uint64_t, utils::tuple_hash>& task_map) {
            std::vector<cm::pending_compaction> res;
            res.reserve(task_map.size());
            for (auto i : task_map) {
                cm::pending_compaction task;
                task.ks = i.first.first;
                task.cf = i.first.second;
                task.task = i.second;
                res.emplace_back(std::move(task));
            }
            return make_ready_future<json::json_return_type>(res);
        });
    });
```

It's returning a `std::vector<cm::pending_compaction>`, but what turns that into a JSON object with a `task` field?

amnonh:
It's a vector of objects with a task field

penberg:
Aah, here it is:

```c++
struct pending_compaction : public json::json_base {
    /**
     * The column family name
     */
  json::json_element< sstring > cf;

    /**
     * The keyspace name
     */
  json::json_element< sstring > ks;

    /**
     * The number of pending tasks
     */
  json::json_element< long > task;

void register_params() {
  add(&cf,"cf");
  add(&ks,"ks");
  add(&task,"task");

}
```

Thanks for the clarification, @amnonh!

PR #79, opened 2019-08-28
run scylla-jmx in a systemd slice

glommer:
Scylla now supports server-defined systemd slices that are used to provide
isolation between components.

This patch adds scylla-jmx to the helper slice. This will guarantee that
scylla-jmx does not use too much resources, influencing the server performance.

Signed-off-by: Glauber Costa <glauber@scylladb.com>

### DIFF
--- a/dist/common/systemd/scylla-jmx.service
+++ b/dist/common/systemd/scylla-jmx.service
@@ -11,6 +11,7 @@ Group=scylla
 ExecStart=/opt/scylladb/jmx/scylla-jmx $SCYLLA_JMX_PORT $SCYLLA_API_PORT $SCYLLA_API_ADDR $SCYLLA_JMX_ADDR $SCYLLA_JMX_FILE $SCYLLA_JMX_LOCAL $SCYLLA_JMX_REMOTE $SCYLLA_JMX_DEBUG
 KillMode=process
 Restart=on-abnormal
+Slice=scylla-helper.slice
 
 [Install]
 WantedBy=multi-user.target


glommer:
Based on what do you say this will enforce a minimum?
The helper slice only specifies a maximum, which is percentage based.

slivne:
I think the JVM will not be able to boot since the maximum is enforced

Scylla needs to operate in developer mode on 1 G machine - I think with
these changes it will not

glommer:
Shlomi,

For small machines we override the reservation during post install. I tested that specifically (although they were not 1GB).

However the post install script overrides scylla-helper.slice and sets the MemoryMax to 1.2GB (which will effectively means it is unlimited for such small machines).

Once more, I don't understand your opposition.

glommer:
@slivne I argued above that this will work on small machines because for those, we lift the reservation.

Do you have any concrete proof that it won't?

glommer:
@slivne ping

slivne:
Sorry Ok I missed the post_install script that changes the helper.slice memory limit.

So it seems to be ok at least to me

glommer:
News?

glommer:
force pushed.

glommer:
Folks, what is missing here?

I'd like to see this as part of 3.2. This is a one line patch, and is in PR state for 2 months

dorlaor:
:(

glommer:
Apparently this got conflicts again, so I will force push it once more.
This is a one line patch that is sitting here since August and that I had to rebase twice, with no changes.

If this doesn't get merged until the end of the week I am closing the PR

glommer:
@slivne force-pushed.

PR #83, opened 2019-11-06
HTTP client per instance

### LINKED ISSUE #82: Memory Leak due to Client per request
According to the docs for [javax.ws.rs.client.Client](https://jax-rs.github.io/apidocs/2.0.1/javax/ws/rs/client/Client.html):

> Clients are heavy-weight objects that manage the client-side communication infrastructure. Initialization as well as disposal of a Client instance may be a rather expensive operation. It is therefore advised to construct only a small number of Client instances in the application. Client instances must be properly closed before being disposed to avoid leaking resources.

The current code creates this heavy-weight object per API Request:

https://github.com/scylladb/scylla-jmx/blob/branch-3.1/src/main/java/com/scylladb/jmx/api/APIClient.java#L90-L101:

```java
    public Invocation.Builder get(String path, MultivaluedMap<String, String> queryParams) {
        Client client = ClientBuilder.newClient(new ClientConfig());
        WebTarget webTarget = client.target(getBaseUrl()).path(path);
        if (queryParams != null) {
            for (Entry<String, List<String>> qp : queryParams.entrySet()) {
                for (String e : qp.getValue()) {
                    webTarget = webTarget.queryParam(qp.getKey(), e);
                }
            }
        }
        return webTarget.request(MediaType.APPLICATION_JSON);
    }
```

ambantis:
- Create javax HTTP client once per instance instead of per request.

This PR resolves #82 for branch 3-1

### DIFF
--- a/src/main/java/com/scylladb/jmx/api/APIClient.java
+++ b/src/main/java/com/scylladb/jmx/api/APIClient.java
@@ -41,7 +41,19 @@ import org.glassfish.jersey.client.ClientConfig;
 import com.scylladb.jmx.utils.SnapshotDetailsTabularData;
 
 public class APIClient {
-    private Map<String, CacheEntry> cache = new HashMap<String, CacheEntry>();
+    private final Map<String, CacheEntry> cache;
+    private final APIConfig config;
+    private final Client client;
+    private final JsonReaderFactory factory;
+
+    private static final Logger logger = Logger.getLogger(APIClient.class.getName());
+
+    public APIClient(APIConfig config) {
+        this.factory = Json.createReaderFactory(null);
+        this.cache = new HashMap<String, CacheEntry>();
+        this.config = config;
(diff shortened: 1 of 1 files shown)


amnonh:
@elcallio can you please have a look?

elcallio:
LGTM

penberg:
Hi @ambantis, thanks for the fix! We need to merge this to `master` first and only then backport the fix to 3.1 to ensure that future versions won't regress. Please either (1) rebase the fix on top of `master` or (2) let me know if you prefer me to do it for you.

ambantis:
@penberg, I'm using that branch as a patched version of the current jmx jar so would prefer to leave that branch as-is, but I'll create a new PR that is based upon master branch

ambantis:
closing this PR in favor of #86, which is based on master branch

PR #86, opened 2019-11-14
HTTP Client per instance.

ambantis:
- Create javax HTTP client once per instance instead of per request.

This PR resovles #82 for master branch.


amnonh:
it was reviewed and is waiting to be merged, there was an issue with the base branch but @penberg can out-smart git

PR #89, opened 2019-12-30
dist: drop symlink to scripts

syuu1228:
This is scylla-jmx part of https://github.com/scylladb/scylla/pull/5530

After we stopped replacing /usr/lib/scylla with symlink,
creating symlink on /opt/scylladb/scripts/jmx become meaningless,
so we can drop it now.

We able to introduce new symlink on /usr/lib/scylla, but it likely no user
directly invoke scripts under /usr/lib/scylla/jmx, so we don't have to
do that.

### DIFF
--- a/dist/redhat/scylla-jmx.spec.mustache
+++ b/dist/redhat/scylla-jmx.spec.mustache
@@ -60,7 +60,6 @@ rm -rf $RPM_BUILD_ROOT
 /opt/scylladb/jmx/scylla-jmx
 /opt/scylladb/jmx/scylla-jmx-1.0.jar
 /opt/scylladb/jmx/symlinks/scylla-jmx
-/opt/scylladb/scripts/jmx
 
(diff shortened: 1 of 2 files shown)


penberg:
@syuu1228 You say that "After we stopped replacing /usr/lib/scylla with symlink [...]", but in which commit and repository did we do this?

penberg:
Aah, now I understand, we did not do it yet, but it's the pull request https://github.com/scylladb/scylla/pull/5530 that removes the symlink...

penberg:
The pull request looks good to me after https://github.com/scylladb/scylla/pull/5530 is merged.

penberg:
Merged as commit https://github.com/scylladb/scylla-jmx/commit/4c8660d41ad5a3c1157711a375b20af6ae28fa1d, not sure why Github didn't auto-close this.

PR #95, opened 2020-01-16
dist/redhat: override systemd macros to CentOS7 version

### LINKED ISSUE #94: dist/redhat: systemd macro still not same with CentOS7 one
As @avikivity pointed out at https://github.com/scylladb/scylla-jmx/pull/92#discussion_r363678653, Fedora 31 version of systemd macros does not work correctly on CentOS7, since CentOS7 does not support "file trigger" feature.
Even after https://github.com/scylladb/scylla-jmx/commit/05d437893e07eecc8b946186217b71b13e155f29, scriptlets on old scylla .rpm and new scylla .rpm is not completely same (attached diff)

I think it's better to align with CentOS7 macro.
```
--- /home/syuu/scriptlet.enterprise     2020-01-17 00:57:57.186990306 +0900
+++ /home/syuu/scriptlet.oss    2020-01-17 00:58:04.932009972 +0900
@@ -12,17 +12,16 @@
 fi
 postinstall scriptlet (using /bin/sh):

-if [ $1 -eq 1 ] ; then
+
+if [ $1 -eq 1 ] && [ -x /usr/bin/systemctl ] ; then
         # Initial installation
-        systemctl preset scylla-jmx.service >/dev/null 2>&1 || :
+        /usr/bin/systemctl --no-reload preset scylla-jmx.service || :
 fi
 preuninstall scriptlet (using /bin/sh):

-if [ $1 -eq 0 ] ; then
+
+if [ $1 -eq 0 ] && [ -x /usr/bin/systemctl ] ; then
         # Package removal, not upgrade
-        systemctl --no-reload disable scylla-jmx.service > /dev/null 2>&1 || :
-        systemctl stop scylla-jmx.service > /dev/null 2>&1 || :
+        /usr/bin/systemctl --no-reload disable --now scylla-jmx.service || : 
 fi
-postuninstall scriptlet (using /bin/sh):
-
-systemctl daemon-reload >/dev/null 2>&1 || :
+postuninstall program: /bin/sh
```

syuu1228:
Fedora 31 version of systemd macros does not work correctly on CentOS7,
since CentOS7 does not support "file trigger" feature.
Even after 05d4378, scriptlets on old scylla .rpm and new scylla .rpm is
not completely same.

To fix the issue we need to override macros with CentOS7 version.

Fixes #94

### DIFF
--- a/dist/redhat/build_rpm.sh
+++ b/dist/redhat/build_rpm.sh
@@ -69,6 +69,7 @@ RPMBUILD=$(readlink -f ../)
 mkdir -p $RPMBUILD/{BUILD,BUILDROOT,RPMS,SOURCES,SPECS,SRPMS}
 
 ln -fv $RELOC_PKG $RPMBUILD/SOURCES/
+ln -fv dist/redhat/systemd.inc $RPMBUILD/SOURCES/
 pystache dist/redhat/scylla-jmx.spec.mustache "{ \"version\": \"$SCYLLA_VERSION\", \"release\": \"$SCYLLA_RELEASE\", \"product\": \"$PRODUCT\", \"$PRODUCT\": true }" > $RPMBUILD/SPECS/scylla-jmx.spec
 
 # this rpm can be install on both fedora / centos7, so drop distribution name from the file name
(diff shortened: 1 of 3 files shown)


syuu1228:
Pushed re-implemented version based on @avikivity 's suggestion on scylladb/scylla#5605

syuu1228:
@avikivity ping, it's jmx part of https://github.com/scylladb/scylla/pull/5605

avikivity:
I dequeued it since scylladb/scylla#5605 was also dequeued.

PR #96, opened 2020-02-27
install.sh: add dependency check and postinst script for manual install

syuu1228:
To install scylla-jmx using install.sh easily, we need:
 - run dependency check before install
 - run postinst script after install

But we don't want to run them when we build .rpm/.deb package,
we also need to add --packaging option to skip them.

See scylladb/scylla#5830


syuu1228:
@avikivity this is scylla-jmx part of merged PR scylladb/scylla#5873 pending for few months, please review

PR #99, opened 2020-04-01
StorageService: Add the scrub 3.11 command implementation

amnonh:
The scrub command was not supported from node_tool, but now when we want
to enable it the current API is not compatible with the 3.11 MBean
definition.

This patch adds the definition to the MBean and the implementation to
StorageService.

It also addresses two problems with the old scrub implementation, just
in case someone will use them.

1. Implementation didn't pass the parameters to the API.
2. A stub implementation called itself instead of calling an actual
implementation.

This patch will enable to test the command from nodetool additional
changes may come on top of it if more command line options will be
supported.

Relates to scylladb/scylla#6072


slivne:
@penberg ping

slivne:
@penberg ping

slivne:
@penberg ping

PR #103, opened 2020-05-11
dist/redhat: drop dependency on pystache

syuu1228:
Same as https://github.com/scylladb/scylla/pull/6313, drop dependency on
pystache since it nolong present in Fedora 32.


penberg:
Queued as 85cbf486d0ff237171b45d347a44553adf47d38b.

PR #113, opened 2020-06-07
dist/debian: don't install systemd unit by install.sh, use debian/*.service

syuu1228:
Installing *.service by install.sh script causes the error on installing .deb
package, use debian/*.service instead.

Fixes scylladb/scylla#6010
Related scylladb/scylla#5640
Related https://github.com/scylladb/scylla/commit/29285b28e2de2f07593a65a5f03977da440b2e42

### DIFF
--- a/dist/debian/debian/rules
+++ b/dist/debian/debian/rules
@@ -8,7 +8,7 @@ override_dh_auto_clean:
 
 override_dh_auto_install:
 	dh_auto_install
-	cd scylla-jmx; ./install.sh --root "$(CURDIR)/debian/$(DEB_SOURCE)" --sysconfdir /etc/default
+	cd scylla-jmx; ./install.sh --root "$(CURDIR)/debian/tmp" --sysconfdir /etc/default
 
 override_dh_installinit:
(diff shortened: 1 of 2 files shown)


avikivity:
I don't see scylla-jmx.service. Did you refer to https://manpages.debian.org/testing/debhelper/dh_systemd_enable.1.en.html when you wrote debian/.service?

syuu1228:
@avikivity scylla-jmx.service is already placed at debian/scylla-jmx.service on .deb building time:
https://github.com/scylladb/scylla-jmx/blob/master/dist/debian/debian_files_gen.py#L56
What currently we facing is, we package scylla-jmx.service both in debian/scylla-jmx/ and debian/*.service, it seems like the .deb package tries to install the file twice and causes the error.

It probably able to resolve by dropping debian/*.service and adding systemd unit file handling script on debian/*.postinst and debian/*.postrm manually (it basically just running systemctl daemon-reload), but I think current way is more simpler than that.

avikivity:
Ah, thanks for the explanation.

PR #118, opened 2020-06-24
MetricsRegistry: Support both types of histograms

### LINKED ISSUE #117: Scylla JMX should accept both Histogram formats
There are two type of histogram, the older was deprecated a long time ago.

In MetricsRegistry JmxHistogram already support both types.
The JMXTimer should support that as well

amnonh:
There are two types of histograms, the older deprecated histogram and
the new estimated_histogram.

This patch changes the implementation of MetricsRegistry so the JmxTimer
would support both types, similiar to JmxHistogram.

Fixes #117

### DIFF
--- a/src/main/java/org/apache/cassandra/metrics/MetricsRegistry.java
+++ b/src/main/java/org/apache/cassandra/metrics/MetricsRegistry.java
@@ -654,7 +654,11 @@ public class MetricsRegistry {
         public void update(JsonObject obj) {
             // TODO: this is not atomic.
             super.update(obj.getJsonObject("meter"));
-            histogram = new Histogram(obj.getJsonObject("hist"));
+            if (obj.containsKey("hist")) {
+                histogram = new Histogram(obj.getJsonObject("hist"));
+            } else if (obj.containsKey("buckets")) {
+                histogram = new Histogram(new EstimatedHistogram(obj));
+            }
         }
 
         @Override


src/main/java/org/apache/cassandra/metrics/MetricsRegistry.java
```diff
@@ -654,7 +654,11 @@ public JmxTimer(String url, long interval) {
         public void update(JsonObject obj) {
             // TODO: this is not atomic.
             super.update(obj.getJsonObject("meter"));
-            histogram = new Histogram(obj.getJsonObject("hist"));
+            if (obj.containsKey("hist")) {
+                histogram = new Histogram(obj.getJsonObject("hist"));
+            } else if (obj.containsKey("buckets")) {
+                histogram = new Histogram(new EstimatedHistogram(obj));
+            }
```
penberg, line 661: The logic in `JmxHistogram` has an additional `else` path:

```java
        @Override
        public void update(JsonObject obj) {
            if (obj.containsKey("hist")) {
                obj = obj.getJsonObject("hist");
            }
            if (obj.containsKey("buckets")) {
                histogram = new Histogram(new EstimatedHistogram(obj));
            } else {
                histogram = new Histogram(obj);
            }
        }
```

Which one of the two is correct?

amnonh: it's mutually exclusive (it's either hist or buckets) but I'm closing this PR because I will take a different path

amnonh:
I'm going to take a different path with the implementation

PR #119, opened 2020-06-24
gitignore: fix typo and add scylla-apiclient/target/

bhalevy:
when building scylla-jmx in place `scylla-apiclient/target/` is left behind and should be ignored by `.gitignore` otherwise the scylla submodule directory appears to be dirty.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 /target/
 /bin/
 dependency-reduced-pom.xml
+scylla-apiclient/target/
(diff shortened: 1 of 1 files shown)


bhalevy:
@amnonh / @avikivity can you please review and merge?

penberg:
Queued as https://github.com/scylladb/scylla-jmx/commit/4727910b5e9b59b79774d4ee4cb335056c25b4d2

PR #122, opened 2020-07-14
dist: redhat: reduce log spam from unpacking sources when building rpm

avikivity:
rpmbuild defaults to logging the name of every file it unpacks from
the archive. This is quite a lot for Java applications.

Make it quiet with the %setup -q flag.


penberg:
Queued as aa94fe53e06335549ba281cd207c017d1378e4dd

PR #125, opened 2020-08-02
dist: debian: support non-x86

avikivity:
The package is already arch independent, so remove the artifical
restriction to x86.


penberg:
Queued as c5ed83178ada5f0633184ea226d733055e229a25.

PR #128, opened 2020-08-24
dist/debian/control.template:remove conflict tag for java 11 and add it as dependency 

yaronkaikov:
Today before installing jmx we must have java-8 installed.
In Debian10 the default java is 11, Removing the conflict tag and adding java 11 to dependencies list


penberg:
@yaronkaikov The pull request had an extra commit and the _commit_ lacked description. I fixed those up and queued the fix as commit 23da40b559ed096074995ca97c136363ca5c265d. Thanks!

Issue #129, opened 2020-08-25
scylla-jmx failed to start after offline installation


amoskong:
version: unified-package-0.20200824.9636a3399.tar.gz

#### Install steps:

- curl http://downloads.scylladb.com.s3.amazonaws.com/relocatable/unstable/master/2020-08-24T14%3A58%3A54Z/scylla-unified-package-0.20200824.9636a3399.tar.gz -o u.tar.gz
- tar xvfz u.tar.gz
- ./install.sh --nonroot --prefix `pwd`/install_root

#### start scylla

systemctl --user start scylla-server

#### systemctl --user status scylla-jmx -f |less
```
● scylla-jmx.service - Scylla JMX
   Loaded: loaded (/home/scylla-test/.config/systemd/user/../../../install_root/etc/systemd/scylla-jmx.service; linked; vendor preset: enabled)
  Drop-In: /home/scylla-test/.config/systemd/user/scylla-jmx.service.d
           └─nonroot.conf
   Active: failed (Result: exit-code) since Tue 2020-08-25 22:19:49 UTC; 25s ago
  Process: 48818 ExecStart=/home/scylla-test/install_root/jmx/scylla-jmx $SCYLLA_JMX_PORT $SCYLLA_API_PORT $SCYLLA_API_ADDR $SCYLLA_JMX_ADDR $SCYLLA_JMX_FILE $SCYLLA_JMX_LOCAL $SCYLLA_JMX_REMOTE $SCYLLA_JMX_DEBUG (code=exited, status=200/CHDIR)
 Main PID: 48818 (code=exited, status=200/CHDIR)

Aug 25 22:19:49 artifacts-centos8-jenkins-db-node-b7a4fdf8-0-1 systemd[5910]: Started Scylla JMX.
Aug 25 22:19:49 artifacts-centos8-jenkins-db-node-b7a4fdf8-0-1 systemd[48818]: scylla-jmx.service: Changing to the requested working directory failed: No such file or directory
Aug 25 22:19:49 artifacts-centos8-jenkins-db-node-b7a4fdf8-0-1 systemd[48818]: scylla-jmx.service: Failed at step CHDIR spawning /home/scylla-test/install_root/jmx/scylla-jmx: No such file or directory
Aug 25 22:19:49 artifacts-centos8-jenkins-db-node-b7a4fdf8-0-1 systemd[5910]: scylla-jmx.service: Main process exited, code=exited, status=200/CHDIR
[...]

PR #131, opened 2020-08-26
fix offline installation

amoskong:
Currently after offline installation, the scylla-jmx fails to be started.
This patch fixed the problem.

Fixes: https://github.com/scylladb/scylla/issues/7098 by [PATCH] install.sh: check both openjdk-8 and openjdk-11

Fixes: https://github.com/scylladb/scylla-jmx/issues/129 by [PATCH] nonroot.conf: set WorkingDirectory to empty

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -78,12 +78,12 @@ if ! $packaging; then
     has_java=false
     if [ -x /usr/bin/java ]; then
         javaver=$(/usr/bin/java -version 2>&1|head -n1|cut -f 3 -d " ")
-        if [[ "$javaver" =~ ^\"1.8.0 ]]; then
+        if [[ "$javaver" =~ ^\"1.8.0 || "$javaver" =~ ^\"11.0. ]]; then
             has_java=true
         fi
     fi
     if ! $has_java; then
-        echo "Please install openjdk-8 before running install.sh."
+        echo "Please install openjdk-8 or openjdk-11 before running install.sh."
         exit 1
     fi
 fi
--- a/install.sh
+++ b/install.sh
@@ -133,6 +133,7 @@ ExecStart=
 ExecStart=$rprefix/jmx/scylla-jmx \$SCYLLA_JMX_PORT \$SCYLLA_API_PORT \$SCYLLA_API_ADDR \$SCYLLA_JMX_ADDR \$SCYLLA_JMX_FILE \$SCYLLA_JMX_LOCAL \$SCYLLA_JMX_REMOTE \$SCYLLA_JMX_DEBUG
 User=
 Group=
+WorkingDirectory=
 EOS
     if [ ! -d ~/.config/systemd/user/scylla-jmx.service.d ]; then
         mkdir -p ~/.config/systemd/user/scylla-jmx.service.d


amoskong:
Append a patch to fix https://github.com/scylladb/scylla-jmx/issues/132
- [PATCH] nonroot.conf: use customized sysconfdir in env file path

install.sh
```diff
@@ -78,12 +78,12 @@ if ! $packaging; then
     has_java=false
     if [ -x /usr/bin/java ]; then
         javaver=$(/usr/bin/java -version 2>&1|head -n1|cut -f 3 -d " ")
-        if [[ "$javaver" =~ ^\"1.8.0 ]]; then
+        if [[ "$javaver" =~ ^\"1.8.0 || "$javaver" =~ ^\"11.0.8 ]]; then
```
penberg, line 81: Should we relax the version check a bit? IOW, lose that patch level (".8") to avoid updating this all the time as new Java versions are released?

penberg:
@syuu1228 Please review this.

install.sh
amoskong: I considered that, but 1.8.0 is fixed version. Not sure the major version will be change.

However, in https://packages.debian.org/search?keywords=openjdk-11-jre-headless
I saw both 11.0.6 and 11.0.8

Let's change it to 11.0

syuu1228 (commented):
I think we shouldn't apply $sysconfdir = /etc/default on nonroot mode.
Why we want to use it on Debian variants is it's distribution's default sysconfig dir.
On nonroot mode we uses $prefix/$sysconfdir, it's not /etc/default directory anyway, so no reason to apply the directory name here.
I suppose just always use $prefix/etc/sysconfig is fine for nonroot mode.

```
The env file path in nonroot.conf is /home/scylla-test/scylladb/etc/sysconfig/scylla-jmx.
Actually it's installed to /home/scylla-test/scylladb/etc/default/scylla-jmx,
because the special sysconfdir is assigned in install commandline.
```

syuu1228:
Other changes looks okay to me.

amoskong:
> I think we shouldn't apply $sysconfdir = /etc/default on nonroot mode.
> Why we want to use it on Debian variants is it's distribution's default sysconfig dir.
> On nonroot mode we uses $prefix/$sysconfdir, it's not /etc/default directory anyway, so no reason to apply the directory name here.
> I suppose just always use $prefix/etc/sysconfig is fine for nonroot mode.

If sysconfdir isn't assigned,   $prefix/etc/sysconfig  ==  $retc/../$sysconfig
If sysconfdir is assigned, my change will still work.

> ```
> The env file path in nonroot.conf is /home/scylla-test/scylladb/etc/sysconfig/scylla-jmx.
> Actually it's installed to /home/scylla-test/scylladb/etc/default/scylla-jmx,
> because the special sysconfdir is assigned in install commandline.
> ```

OK, I got it. You expect the syscondir is only assigned for root install of debian-like distros.
It should not be assigned for nonroot install of debian-like distros.
The limit should be added to parameters parser of install.sh

I'm fine to use default sysconfdir for nonroot install of debian-like distros. So I will remove the last patch. Then we can merge this PR.

Thanks for your review.

syuu1228 (approved):
LGTM

Issue #135, opened 2020-09-01
Column family and metrics cause a very high object (and name) cound in mbean server. 


elcallio:
For every CF, we bind some 50+ metrics, each under two different names (Table and ColumnFamily).
This leads to rather high object counts if the associated server has many CF:s. 

Since objects are also only reaped when a wildcard query is issued, it is somewhat possible to cause 
objects to remain bound that represent dropped CF:s for quite some time, depending on query pattern
or JMX commands run.

PR #136, opened 2020-09-01
JMX footprint work

### LINKED ISSUE #133: TableMetricsObjectname is not serializable
It says it is, but is not. APIMBeanServer replaces it in RPC calls, but it is still occasionally seen in Notifications, which causes exceptions in the JMX connector.

### LINKED ISSUE #134: Column family beans refresh is not syncronized
Later JDKs/connections to mbean server can cause multiple threads to call the registration check code simultaneously. While this is not fatal at all, it will cause more object storms and occasionally higher object pressure than needed. 

Should use a lightweight sync and try to avoid unnecessary repeated calls to the API server.

elcallio:
Fixes #133
Fixes #134 
Refs #135

Makes CF mbean refresh code synchronized and tries to remove reductant calls if we contend. 
Adds background reaping of dead objects to reduce memory load in (test) scenarios where we manage
to refresh to add, but not cause removal (i.e. no wildcard queries). 

TableMetricsObjectName serialization is fixed in the series because without it we see loads of 
exceptions when refreshing the mbean set.

### DIFF
--- a/src/main/java/com/scylladb/jmx/metrics/RegistrationChecker.java
+++ b/src/main/java/com/scylladb/jmx/metrics/RegistrationChecker.java
@@ -1,0 +1,69 @@
+package com.scylladb.jmx.metrics;
+
+import static com.scylladb.jmx.metrics.RegistrationMode.Remove;
+import static com.scylladb.jmx.metrics.RegistrationMode.Wait;
+import static java.util.EnumSet.allOf;
+import static java.util.EnumSet.of;
+
+import java.net.UnknownHostException;
+import java.util.EnumSet;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
+
+import javax.management.OperationsException;
+
+import com.scylladb.jmx.api.APIClient;
+import com.sun.jmx.mbeanserver.JmxMBeanServer;
+
+/**
+ * Helper type to do optional locking for registration. Allows for
+ * per-bind-point locks and registration, instead of per-type or per-instance
+ * locks which may be misguiding, since for example one instance can be bound to
+ * many MBeanServers etc.
+ * 
+ * Also allows for polled checks, i.e. try-lock and either wait or skip. Wait,
+ * because we probably should not repeat things hidden by this type too often,
+ * and skip because for example a periodic task checking can just skip if a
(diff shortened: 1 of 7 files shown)


src/main/java/org/apache/cassandra/metrics/TableMetrics.java
```diff
@@ -400,6 +401,18 @@ public boolean isPropertyListPattern() {
...
+         * This type is not really serializable. 
+         * Replace it with vanilla objectname.
+         */
+		private Object writeReplace() throws ObjectStreamException {
+			try {
+				return new ObjectName(getDomain(), getKeyPropertyList());
+			} catch (MalformedObjectNameException e) {
+				throw new InvalidObjectException(toString());
+			}
+		}
```
penberg, line 415: Did you discover the one true indentation style -- tabs?

src/main/java/com/scylladb/jmx/metrics/RegistrationChecker.java
```diff
@@ -0,0 +1,53 @@
...
+		check(client, server, REMOVE_NO_WAIT);
+	}
+
+	public final void check(APIClient client, JmxMBeanServer server) throws OperationsException, UnknownHostException {
+		check(client, server, ADD_AND_REMOVE);
+	}
+
+	public final void check(APIClient client, JmxMBeanServer server, EnumSet<RegistrationMode> mode)
+			throws OperationsException, UnknownHostException {
+		if (!lock.tryLock() && mode.contains(Wait)) {
```
penberg, line 35: If `tryLock` fails, but we're not allowed to wait, what is supposed to happen? Right now, we perform `doCheck` without the lock (which I assume is safe?), but we also call `unlock()`...

penberg (commented):
Can you also say few words on why we need `RegistrationChecker`? Why do we need light-weight synchronization for this, instead of slapping `synchronized` everywhere?

src/main/java/org/apache/cassandra/metrics/TableMetrics.java
elcallio: Meh. Ran a new eclipse. We should check the formatter in with the maven project.

elcallio:
Fixed the above notes.

PR #137, opened 2020-09-13
dist: do not install build dependencies on build script

syuu1228:
We do not want to install dependencies on package building time,
we want to install it on dbuild container.
So drop package installation from the scripts.

scylladb/scylla#7219


penberg:
@syuu1228 For future reference, we no longer use the `next` branch in `scylla-jmx-git`. @avikivity Perhaps we ought to just remove it now?

Anyway, I fixed up the merge conflict and merged this as eb9e58347c0476f3ea59f9af87653ca852a199f0.

penberg:
@syuu1228 I removed this from `master` because it breaks the build. See the discussion in https://github.com/scylladb/scylla-tools-java/pull/203 for more details.

PR #143, opened 2020-10-05
StorageService.java: Use the endpoint for getRangeToEndpointMap

### LINKED ISSUE #36: no support for StorageServiceMBean.getRangeToEndpointMap
We are missing support for this method

Some third party rely on it such as cassandra-reaper

amnonh:
After implementing range_to_endpoint_map endpoint update the API call to
it.

Fixes #36


penberg:
@amnonh Is the API side implemented in `scylla.git` already? Which commit(s)?

amnonh:
The unimplemented API is there (i.e. you will get an unimplemented exception just like the situation today).

The actual implementation is in the PR https://github.com/scylladb/scylla/pull/7329

penberg:
@amnonh Thanks. I guess I grepped for the wrong thing in `scylla.git` because I thought `range_to_endpoint_map` did not exist at all...

PR #144, opened 2020-10-05
install.sh: show warning nonroot mode when systemd does not support user mode

syuu1228:
On older distribution such as CentOS7, it does not support systemd user mode.
On such distribution nonroot mode does not work, show warning message and
skip running systemctl --user.

See scylladb/scylla#7071

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -74,6 +74,11 @@ while [ $# -gt 0 ]; do
     esac
 done
 
+check_usermode_support() {
+    user=$(systemctl --help|grep -e '--user')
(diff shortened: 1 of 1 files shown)


syuu1228:
This is scylla-jmx part of https://github.com/scylladb/scylla/pull/7187

syuu1228:
ping

avikivity:
Actually it was already merged in e3a381d5a128ba0fc94130ae76ca14cfcb644d95, github didn't notice.

PR #146, opened 2020-10-12
Relocatable Package: create product prefixed relocatable archive

eliransin:
The build system was hardcoded to produce a package that is
prefixed with scylla instead of the product name. This is not
in line with out CI system requirements and can be also a source
for confusion.
This commit make the packaging system generate a package of
the format: {product}-jmx-package.tar.gz instead of
scylla-jmx-package.tar.gz

### DIFF
--- a/dist/redhat/scylla-jmx.spec
+++ b/dist/redhat/scylla-jmx.spec
@@ -6,7 +6,7 @@ Group:          Applications/Databases
 
 License:        AGPLv3
 URL:            http://www.scylladb.com/
-Source0:        scylla-jmx-package.tar.gz
+Source0:        %{reloc_pkg}
 
 BuildArch:      noarch
 BuildRequires:  systemd-units
--- a/reloc/build_reloc.sh
+++ b/reloc/build_reloc.sh
@@ -33,6 +33,11 @@ while [ $# -gt 0 ]; do
     esac
 done
 
+VERSION=$(./SCYLLA-VERSION-GEN ${VERSION_OVERRIDE:+ --version "$VERSION_OVERRIDE"})
+# the former command should generate build/SCYLLA-PRODUCT-FILE and some other version
+# related files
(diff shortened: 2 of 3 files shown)


yaronkaikov (approved):

syuu1228:
I think we currently don't do this on scylla main repo and other submodules (we always uses "scylla-" prefix for relocatable packages).
@avikivity @penberg Do we want to rename them? If so I'll send PR for other repos to do same renaming.

eliransin:
> @avikivity @penberg Do we want to rename them? If so I'll send PR for other repos to do same renaming.

@syuu1228 I already have PRS for all other repos.
In addition I have a PR for the scylla repo, I still didn't post it because if I do and the merges in all submodules and the OSS
to enterprise are not close together, we will break scylla enterprise.
The order of things should be:
1. Merge rename to all submodules
2. Merge all submodules to enterprise modules
3. Merge PR in OSS scylla (possibly followed by an update to the submodules)
4. Merge scylla OSS to enterprise imediately followed by a submodule update. Or scylla enterprise will break.

eliransin:
@penberg ping (merge or comment :slightly_smiling_face: )

eliransin:
@scylladb/scylla-jmx-maint can someone merge please?

nyh:
@avikivity @penberg I want to merge this, but discovered that scylla-jmx is missing a "next" branch - it has a next-4.2 etc branch, but not "next". Why? Are we pushing this repository directly to master?

eliransin:
> @avikivity @penberg I want to merge this, but discovered that scylla-jmx is missing a "next" branch - it has a next-4.2 etc branch, but not "next". Why? Are we pushing this repository directly to master?

@nyh this is intentional, this is because those are sub-modules in scylla now so they don't need an individual promotion process,
if something is not right then scylla will not advance it's pointers until the submodule is stable.

PR #150, opened 2020-12-28
install.sh: fix hardcode sysconfdir in EnvironmentFile path

amoskong:
Currently we use fixed sysconfdir name 'sysconfig', it works as
expected. But let's change it to use the assigned sysconfdir, it
still works when a different sysconfdir is assigned.

Signed-off-by: Amos Kong <amos@scylladb.com>

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -104,12 +104,12 @@ fi
 rprefix=$(realpath -m "$root/$prefix")
 if ! $nonroot; then
     retc="$root/etc"
-    rsysconfdir="$root/$sysconfdir"
+    rsysconfdir=$(realpath -m "$root/$sysconfdir")
     rusr="$root/usr"
     rsystemd="$rusr/lib/systemd/system"
 else
     retc="$rprefix/etc"
-    rsysconfdir="$rprefix/$sysconfdir"
+    rsysconfdir=$(realpath -m "$rprefix/$sysconfdir")
     rsystemd="$HOME/.config/systemd/user"
 fi
 
@@ -133,7 +133,7 @@ else
     cat << EOS > "$rsystemd"/scylla-jmx.service.d/nonroot.conf
 [Service]
 EnvironmentFile=
-EnvironmentFile=$retc/sysconfig/scylla-jmx
+EnvironmentFile=$rsysconfdir/scylla-jmx
 ExecStart=
 ExecStart=$rprefix/jmx/scylla-jmx \$SCYLLA_JMX_PORT \$SCYLLA_API_PORT \$SCYLLA_API_ADDR \$SCYLLA_JMX_ADDR \$SCYLLA_JMX_FILE \$SCYLLA_JMX_LOCAL \$SCYLLA_JMX_REMOTE \$SCYLLA_JMX_DEBUG
 User=


install.sh
```diff
@@ -104,12 +104,12 @@ fi
 rprefix=$(realpath -m "$root/$prefix")
 if ! $nonroot; then
     retc="$root/etc"
-    rsysconfdir="$root/$sysconfdir"
+    rsysconfdir=$(realpath -m "$root/$sysconfdir")
```
avikivity, line 107: This is a little dangerous. If /etc/sysconfig is a symlink on the build system (when building an rpm/deb), it will translate that symlink on the build host, even though the translation may be incorrect for the installed system.

install.sh
```diff
@@ -133,7 +133,7 @@ else
     cat << EOS > "$rsystemd"/scylla-jmx.service.d/nonroot.conf
 [Service]
 EnvironmentFile=
-EnvironmentFile=$retc/sysconfig/scylla-jmx
+EnvironmentFile=$rsysconfdir/scylla-jmx
```
avikivity, line 136: Better to execute the translation here.

install.sh
amoskong: You are right. I will remove this change

install.sh
amoskong: I had moved the translation here. Thx.

The translation only cleaned a redundant backslash

```
# Without this fix:
EnvironmentFile=/home/scylla-test/scylladb/etc/sysconfig//scylla-jmx

# Applied this fix:
EnvironmentFile=/home/scylla-test/scylladb/etc/sysconfig/scylla-jmx
```

syuu1228:
Oh sorry it seems like why we requires this change is because I haven't synced changes from scylla repo to scylla-jmx repo.
Anyway, I think we align the way to implement this between scylla repo and scylla-jmx repo.
On scylla repo, we do:
```
        cat << EOS > "$rsystemd"/scylla-server.service.d/nonroot.conf
[Service]
EnvironmentFile=
EnvironmentFile=$rsysconfdir/scylla-server
```
https://github.com/scylladb/scylla/blob/master/install.sh#L372
So scylla-jmx repo should be same.
If current scylla repo's implementation has problem, then replace with new one on both repos, not just one.

amoskong:
> Oh sorry it seems like why we requires this change is because I haven't synced changes from scylla repo to scylla-jmx repo.
> Anyway, I think we align the way to implement this between scylla repo and scylla-jmx repo.
> On scylla repo, we do:
> 
> ```
>         cat << EOS > "$rsystemd"/scylla-server.service.d/nonroot.conf
> [Service]
> EnvironmentFile=
> EnvironmentFile=$rsysconfdir/scylla-server
> ```
> 
> https://github.com/scylladb/scylla/blob/master/install.sh#L372
> So scylla-jmx repo should be same.
> If current scylla repo's implementation has problem, then replace with new one on both repos, not just one.

Currently implement in scylla_repo is fine, a redundant backslash is harmless.

```
# Current scylla_repo::
EnvironmentFile=/home/scylla-test/scylladb/etc/sysconfig//scylla-server

# Fixed scylla-jmx:
EnvironmentFile=/home/scylla-test/scylladb/etc/sysconfig/scylla-jmx
```

So it's fine to only merge this PR.

penberg:
@amoskong Even if `scylla.git` works fine, let's make sure the two solutions are the same. So please open a pull request in `scylla.git` to switch to also use `realpath` and let's merge both.

amoskong:
> @amoskong Even if `scylla.git` works fine, let's make sure the two solutions are the same. So please open a pull request in `scylla.git` to switch to also use `realpath` and let's merge both.

Thanks for the suggestion, I had sent a PR: https://github.com/scylladb/scylla/pull/7860

avikivity:
@syuu1228 are you happy with this?

penberg:
@syuu1228 ping

syuu1228 (approved):
LGTM

PR #157, opened 2021-02-15
storage_service: Add a generic toppartitions endpoint

StarostaGit:
As part of making the toppartitions API more generic
(i.e. being able to consider multiple tables
and keyspaces specified by the user) this commit adds
a JMX endpoint to call the generic Scylla REST API
introduced in [#7864](https://github.com/scylladb/scylla/pull/7864). It has been put inside
storage_service as being now able to query more than
one column family makes it no longer suitable for the
'column_family' group.

Fixes #4520

### DIFF
--- a/src/main/java/org/apache/cassandra/service/StorageService.java
+++ b/src/main/java/org/apache/cassandra/service/StorageService.java
@@ -54,7 +54,14 @@ import javax.management.NotificationBroadcasterSupport;
 import javax.management.NotificationFilter;
 import javax.management.NotificationListener;
 import javax.management.openmbean.CompositeData;
+import javax.management.openmbean.CompositeDataSupport;
+import javax.management.openmbean.CompositeType;
+import javax.management.openmbean.OpenDataException;
+import javax.management.openmbean.OpenType;
+import javax.management.openmbean.SimpleType;
 import javax.management.openmbean.TabularData;
+import javax.management.openmbean.TabularDataSupport;
+import javax.management.openmbean.TabularType;
 import javax.ws.rs.core.GenericType;
 import javax.ws.rs.core.MultivaluedHashMap;
 import javax.ws.rs.core.MultivaluedMap;
@@ -66,6 +73,7 @@ import com.google.common.base.Joiner;
 import com.scylladb.jmx.api.APIClient;
 import com.scylladb.jmx.metrics.MetricsMBean;
 import com.scylladb.jmx.utils.FileUtils;
+import com.google.common.base.Throwables;
 
 /**
  * This abstraction contains the token/identifier of this node on the identifier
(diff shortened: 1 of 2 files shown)


avikivity:
@amnonh I'm relying on your review here.

src/main/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -1749,4 +1790,33 @@ public int scrub(boolean disableSnapshot, boolean skipCorrupted, boolean checkDa
         APIClient.set_query_param(queryParams, "cf", APIClient.join(columnFamilies));
         return client.getIntValue("/storage_service/keyspace_scrub/" + keyspaceName, queryParams);
     }
+
+    @Override
+    public CompositeData getToppartitions(String sampler, List<String> keyspaceFilters, List<String> tableFilters, int duration, int capacity, int count) throws OpenDataException {
```
amnonh, line 1795: I've comment on the other PR, you can just keep keyspaceFilters as is for the entire way from nodetool to the API

amnonh (changes requested):

src/main/java/org/apache/cassandra/service/StorageService.java
StarostaGit: I don't see any comments about this on the other PRs. What do you mean by keeping those "as is"?

penberg: @amnonh ping

amnonh: In the nodetool/nodeprobe side you take a string and split it into an array, then here you take that array and makes it a string, you can just pass it as it is

StarostaGit: It seems to me that it's more elegant and universal expecting a list, not a string in the jmx endpoint. List is self explanatory, while accepting a specifically separated string requires some form of conveying that information, like documentation or whatever. Also, things outside of jmx do not really need to know how the APIClient inside joins lists for queries - was it to ever change, we then would have to change it also in nodetool and every other place the exposed jmx endpoint is being called. The lists shouldn't be long so performance is not an issue either.

That's my reasoning behind this, but you have the final word. If you prefer it all to just pass the string around, I'll change it.

StarostaGit:
rebased to resolve conflicts
@amnonh ping

amnonh:
LGTM - the only concern I had was about spliting the string and join it back again, but I'm fine with this approach

StarostaGit:
this should be merged after https://github.com/scylladb/scylla/pull/7864 is merged

StarostaGit:
> this should be merged after [scylladb/scylla#7864](https://github.com/scylladb/scylla/pull/7864) is merged

It's merged now, so this can be now committed as well

tzach:
@StarostaGit can you please provide a few example for using nodetool top partitions with the new options?

StarostaGit:
> @StarostaGit can you please provide a few example for using nodetool top partitions with the new options?

@tzach should I put it somewhere or just list it here?
But in general it works as follows:
`nodetool toppartitions` -> considers all partitions in all tables in all keyspaces with default duration
`nodetool toppartitions -d 1000` -> same as above, but the duration is specified to be 1000ms
`nodetool toppartitions --cf-filters ks:t,system:status` -> toppartitions query only for partitions from table t in keyspace ks and table status in keyspace system
`nodetool toppartitions --ks-filters ks,system` -> toppartitions query considering partitions from all tables inside ks and system keyspaces

PR #165, opened 2021-05-10
storage_service: Fix getToppartitions to always return both reads and…

StarostaGit:
… writes

In line with the previous API, the getToppartitions function returned
results for one specified sampler (reads OR writes). This forced
the user to call the function once for each sampler, which is
suboptimal.
This commit changes the signature so that results for both samplers
are returned and the user can then pick whichever they need.

This is part of the fix for #243 and scylladb/scylla#8459

### DIFF
--- a/src/main/java/org/apache/cassandra/service/StorageService.java
+++ b/src/main/java/org/apache/cassandra/service/StorageService.java
@@ -28,6 +28,7 @@ import java.io.IOException;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
 import java.nio.ByteBuffer;
+import java.util.Arrays;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -1798,31 +1800,37 @@ public long getUptime() {
     }
 
     @Override
-    public CompositeData getToppartitions(String sampler, List<String> keyspaceFilters, List<String> tableFilters, int duration, int capacity, int count) throws OpenDataException {
+    public Map<String, CompositeData> getToppartitions(List<String> keyspaceFilters, List<String> tableFilters, int duration, int capacity, int count) throws OpenDataException {
```
avikivity, line 1803: Is it possible to keep the old signature too? So that old nodetool works with new jmx. Keep in mind that it's possible to run nodetool remotely, so the packages may not be in sync.

If jmx matches by full signature, not just name, then it should just work.

avikivity: Also: maybe pass a list of samplers, in case Scylla grows the ability to select samplers. This will remove the need to change the signature once more in the future.

avikivity:
I merged this, will let the tests verify that it still works with the old nodetool. Once it passes we can merge the new nodetool, ping me if I forget.

Can we make the new nodetool work with old jmx? I guess we can with try/catch.

StarostaGit:
@avikivity by old jmx you mean before this commit or the one before generalizing toppartitions?

PR #171, opened 2021-06-25
install.sh: add supervisor support

syuu1228:
Bring supervisor support from dist/docker to install.sh, make it
installable from relocatable package.
This enables to use supervisor with nonroot / offline environment,
and also make relocatable package able to run in Docker environment.

See scylladb/scylla#8849

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -32,6 +32,7 @@ Options:
   --nonroot                shortcut of '--disttype nonroot'
   --sysconfdir /etc/sysconfig   specify sysconfig directory name
   --packaging               use install.sh for packaging
+  --supervisor             enable supervisor to manage scylla processes
   --help                   this helpful message
 EOF
     exit 1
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -156,10 +162,10 @@ if $nonroot; then
...
     sed -i -e "s#/opt/scylladb/jmx#$rprefix/jmx#g" "$rsysconfdir"/scylla-jmx
-    if check_usermode_support; then
+    if ! $supervisor && check_usermode_support; then
         systemctl --user daemon-reload
     fi
     echo "Scylla-JMX non-root install completed."
-elif ! $packaging; then
+elif ! $supervisor && ! $packaging; then
     systemctl --system daemon-reload
 fi
```
avikivity, line 171: I see it doesn't copy any supervisor config files. I think they should be moved to this package, just like the systemd integration is here.

syuu1228:
Closing, we decided not changing this module. See: scylladb/scylla#9142

PR #175, opened 2021-08-18
scrub: support scrub_mode

bhalevy:
Add support for the scrub_mode option for the scylla api.

### DIFF
--- a/src/main/java/org/apache/cassandra/service/StorageServiceMBean.java
+++ b/src/main/java/org/apache/cassandra/service/StorageServiceMBean.java
@@ -351,6 +351,10 @@ public interface StorageServiceMBean extends NotificationEmitter {
      * scrubbed.
      *
      * Scrubbed CFs will be snapshotted first, if disableSnapshot is false
+     *
+     * scrubMode controls what scrub does when encountering corruption.
+     * It replaces skipCorrupted where skipCorrupted is equivalent to scrubMode="SKIP".
+     * Can be one of: "ABORT", "SKIP", "SEGREGATE", or "VALIDATE".
      */
     @Deprecated
     public int scrub(boolean disableSnapshot, boolean skipCorrupted, String keyspaceName, String... tableNames)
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/cassandra/service/StorageServiceMBean.java
```diff
@@ -364,10 +364,15 @@ public int scrub(boolean disableSnapshot, boolean skipCorrupted, boolean checkDa
...
             String... columnFamilies) throws IOException, ExecutionException, InterruptedException;
 
+    @Deprecated
     public int scrub(boolean disableSnapshot, boolean skipCorrupted, boolean checkData, boolean reinsertOverflowedTTL,
             int jobs, String keyspaceName, String... columnFamilies)
             throws IOException, ExecutionException, InterruptedException;
 
+    public int scrub(boolean disableSnapshot, boolean skipCorrupted, String scrubMode, boolean checkData, boolean reinsertOverflowedTTL,
+            int jobs, String keyspaceName, String... columnFamilies)
+            throws IOException, ExecutionException, InterruptedException;
```
denesb, line 374: I think there should be a variant with `skipCorrupted` and one with `scrubMode`, doesn't really make sense to set both, as they conflict (and I don't recall how the conflict is resolved in the API).

bhalevy: Hmm, that makes sense.  Our documentation says that scrubMode trumps skipCorrupted.
https://github.com/scylladb/scylla/blob/f98cb965067b8a8f640840387dcbf043a0885d8c/api/api-doc/storage_service.json#L798

bhalevy: Fixed in v2

bhalevy:
In v2:
- deprecate skipCorrupted, replace the skipCorrupted parameter with scrubMode in new api method.

denesb (approved):

bhalevy:
@scylladb/scylla-jmx-maint please merge

PR #176, opened 2021-08-30
Support `--load-and-stream` option from `nodetool refresh`

jul-stas:
This information is translated to {"load_and_stream", "true"} entry in the
POST request to Scylla's HTTP API at `storage_service/sstables/{keyspace}`
endpoint.

More about this feature: scylladb/scylla#7846
Mutually bonded change: scylladb/scylla-tools-java#273
Refs scylladb/scylla-tools-java#253.

### DIFF
--- a/src/main/java/org/apache/cassandra/service/StorageService.java
+++ b/src/main/java/org/apache/cassandra/service/StorageService.java
@@ -1417,13 +1417,23 @@ public class StorageService extends MetricsMBean implements StorageServiceMBean,
      *            The parent keyspace name
      * @param cfName
      *            The ColumnFamily name where SSTables belong
+     * @param isLoadAndStream
+     *            Whether or not arbitrary SSTables should be loaded (and streamed to the owning nodes)
      */
     @Override
-    public void loadNewSSTables(String ksName, String cfName) {
-        log(" loadNewSSTables(String ksName, String cfName)");
+    public void loadNewSSTables(String ksName, String cfName, boolean isLoadAndStream) {
+        log(" loadNewSSTables(String ksName, String cfName, boolean isLoadAndStream)");
         MultivaluedMap<String, String> queryParams = new MultivaluedHashMap<String, String>();
         queryParams.add("cf", cfName);
+        if (isLoadAndStream) {
+            queryParams.add("load_and_stream", "true");
+        }
         client.post("/storage_service/sstables/" + ksName, queryParams);
+    }
+
+    @Override
+    public void loadNewSSTables(String ksName, String cfName) {
+        loadNewSSTables(ksName, cfName, false);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -1419,10 +1419,13 @@ public void rescheduleFailedDeletions() {
      *            The ColumnFamily name where SSTables belong
      */
     @Override
-    public void loadNewSSTables(String ksName, String cfName) {
+    public void loadNewSSTables(String ksName, String cfName, boolean isLoadAndStream) {
         log(" loadNewSSTables(String ksName, String cfName)");
         MultivaluedMap<String, String> queryParams = new MultivaluedHashMap<String, String>();
         queryParams.add("cf", cfName);
+        if (isLoadAndStream)
```
haaawk, line 1426: Java convention is to always use a {} block after `if`

haaawk (approved):
Is there a way to make sure this works as expected in an automatic way?

nyh:
I guess we can't commit **documentation** with the same PR because it's not in the same repository, but let's at least make sure we have an issue about adding such documentation, and a plan to do it quickly. Otherwise nobody will ever be able to use this new feature.\
I see this new feature is explained in https://groups.google.com/g/scylladb-dev/c/ZIYcEE76BYU?pli=1 but it's missing in https://docs.scylladb.com/operating-scylla/nodetool-commands/refresh/ (and in fact has a NOTE which begs to mention this new feature).

jul-stas:
> 
> Is there a way to make sure this works as expected in an automatic way?

DTest may work - will check. Scylla logs the params of `refresh` queries, so it can be our hook.

src/main/java/org/apache/cassandra/service/StorageService.java
jul-stas: Not in this repo :)
but fixed

jul-stas:
@scylladb/scylla-jmx-maint ping
(dtest is already merged and I promise I'll remember about the docs)

src/main/java/org/apache/cassandra/service/StorageServiceMBean.java
```diff
@@ -808,8 +808,10 @@ default public void setConcurrentCompactors(int value) {
      *            The parent keyspace name
      * @param cfName
      *            The ColumnFamily name where SSTables belong
+     * @param isLoadAndStream
+     *            Whether or not arbitrary SSTables should be loaded (and streamed to the owning nodes)
      */
-    public void loadNewSSTables(String ksName, String cfName);
+    public void loadNewSSTables(String ksName, String cfName, boolean isLoadAndStream);
```
avikivity, line 814: Since this is an RPC interface, doesn't it break compatibility with jmx clients that don't expect it?

jul-stas: Done. (by keeping both ifaces)

PR #178, opened 2021-10-01
removeNode: support ignoreNodes options

bhalevy:
Refs scylladb/scylla-tools-java#225

### DIFF
--- a/src/main/java/org/apache/cassandra/service/StorageService.java
+++ b/src/main/java/org/apache/cassandra/service/StorageService.java
@@ -991,11 +991,19 @@ public class StorageService extends MetricsMBean implements StorageServiceMBean,
      *            the host id to remove
      */
     @Override
-    public void removeNode(String hostIdString) {
-        log(" removeNode(String token)");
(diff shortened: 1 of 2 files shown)


bhalevy:
@asias ping review

src/main/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -991,13 +991,21 @@ public void move(String newToken) throws IOException {
...
         APIClient.set_query_param(queryParams, "host_id", hostIdString);
+        if (ignoreNodes != null) {
+            APIClient.set_query_param(queryParams, "ignore_nodes", ignoreNodes);
+        }
         client.post("/storage_service/remove_node", queryParams);
     }
 
+    public void removeNode(String hostIdString) {
+        String ignoreNodes = null;
+        removeNode(hostIdString, ignoreNodes);
```
haaawk, line 1006: It seems that just passing `null` as a second parameter would be enough, no?

haaawk (approved):

bhalevy:
@scylladb/scylla-maint please merge

PR #179, opened 2021-11-01
StorageService: scrub: fix scrubMode is empty condition

### LINKED ISSUE #180: Testing scrubMode != "" compares references, not values
Change 70b19e6270df1ebe0494af39a2634da0513b38a6 introduce a `scrubMode` parameter to `scrub`.
But, https://github.com/scylladb/scylla-jmx/blob/70b19e6270df1ebe0494af39a2634da0513b38a6/src/main/java/org/apache/cassandra/service/StorageService.java#L1805

misuses Java by comparing `scrubMode` to an empty string using the `!=` operator.
As @haaawk noted, this is wrong as in Java `!=` compares references, not values.

bhalevy:
`!=` compares references not values.

Instead of comparing scrubMode to an empty string, just test !isEmpty().

Fixes #180

### DIFF
--- a/src/main/java/org/apache/cassandra/service/StorageService.java
+++ b/src/main/java/org/apache/cassandra/service/StorageService.java
@@ -1812,7 +1812,7 @@ public class StorageService extends MetricsMBean implements StorageServiceMBean,
             throws IOException, ExecutionException, InterruptedException {
         MultivaluedMap<String, String> queryParams = new MultivaluedHashMap<String, String>();
         APIClient.set_bool_query_param(queryParams, "disable_snapshot", disableSnapshot);
-        if (scrubMode != "") {
+        if (!"".equals(scrubMode)) {
             APIClient.set_query_param(queryParams, "scrub_mode", scrubMode);
         }
         APIClient.set_query_param(queryParams, "cf", APIClient.join(columnFamilies));


src/main/java/org/apache/cassandra/service/StorageService.java
```diff
@@ -1812,7 +1812,7 @@ public int scrub(boolean disableSnapshot, String scrubMode, boolean checkData, b
             throws IOException, ExecutionException, InterruptedException {
         MultivaluedMap<String, String> queryParams = new MultivaluedHashMap<String, String>();
         APIClient.set_bool_query_param(queryParams, "disable_snapshot", disableSnapshot);
-        if (scrubMode != "") {
+        if (!scrubMode.isEmpty()) {
```
haaawk, line 1815: I think `!"".equals(scrubMode)` is safer in case `scrubMode` happens to be a null

haaawk: or check for null explicitly

bhalevy: ok, will fix

haaawk (approved):

bhalevy:
@denesb / @scylladb/scylla-maint please merge

avikivity:
Need issue reference

bhalevy:
> Need issue reference

@avikivity https://github.com/scylladb/scylla-jmx/issues/180

PR #181, opened 2021-11-12
build: replace yum with dnf

avikivity:
dnf has replaced yum on Fedora and CentOS. On modern versions of Fedora,
you have to install an extra package to get the old name working, so
avoid that inconvenience and use dnf directly.


avikivity:
@nyh can you review this please?

nyh (approved):

PR #186, opened 2022-08-22
install.sh: add --without-systemd option

syuu1228:
Since we fail to write files to $USER/.config on Jenkins jobs, we need
an option to skip installing systemd units.
Let's add --without-systemd to do that.

See scylladb/scylla-dtest#2819

----

This is jmx part of scylladb/scylla#11345

### DIFF
--- a/scripts/create-relocatable-package.py
+++ b/scripts/create-relocatable-package.py
@@ -45,9 +45,9 @@ args = ap.parse_args()
 output = args.dest
 
 ar = tarfile.open(output, mode='w|gz')
-# relocatable package format version = 2
+# relocatable package format version = 2.2
 with open('build/.relocatable_package_version', 'w') as f:
-    f.write('2\n')
+    f.write('2.2\n')
 ar.add('build/.relocatable_package_version', arcname='.relocatable_package_version')
 
 pathlib.Path('build/SCYLLA-RELOCATABLE-FILE').touch()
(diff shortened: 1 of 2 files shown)


fruch:
@syuu1228 I think you should advance `.relocatable_package_version` so we can use this in CCM while not breaking backward compatibility

syuu1228:
@fruch incremented .relocatable_package_version to 2.2.

syuu1228:
@avikivity ping

scripts/create-relocatable-package.py
```diff
@@ -45,9 +45,9 @@ def reloc_add(self, name, arcname=None, recursive=True, *, filter=None):
 output = args.dest
 
 ar = tarfile.open(output, mode='w|gz')
-# relocatable package format version = 2
+# relocatable package format version = 2.2
 with open('build/.relocatable_package_version', 'w') as f:
-    f.write('2\n')
+    f.write('2.2\n')
```
avikivity, line 50: Will this require changes elsewhere? Or is it just infomative?

fruch: we will use that in CCM, the whole change is aimed for CCM... (where those systemd files are "escaping" into the user directory, and can cause conflicts in some cases)

avikivity: Is ccm already prepared?

fruch: for this change, no.

we'll do the changes in CCM only after those ones are merge. i.e. it's not a breaking change. (shouldn't be... an scylla CI is covering it, i.e. running dtest)

PR #196, opened 2023-03-26
install.sh: point java to /usr/bin/java

tchaikov:
this should address the regression introduced by
da69e8b2c793af269c23154ee80372faa9955283, as we do pass --package when cooking rpm and deb packages, in which case, $java is not set after that offending commit. and before da69e8b2c7, $java defaults to /usr/bin/java, let's restore that behavior.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -85,6 +85,7 @@ check_usermode_support() {
     [ -n "$user" ]
 }
 
+java=/usr/bin/java
 if ! $packaging; then
     has_java=false
     . /etc/os-release


tchaikov:
@nyh quote from https://github.com/scylladb/scylla-jmx/pull/193#issuecomment-1484029813. all very good questions:

> Ok, I'll merge this PR now, but I'm a bit worried - we don't have CI for this project, and I don't have enough knowledge how it interacts with the other moving pieces - e.g., is it possible that if we change the Java in JMX it will cause problems for dtest? (I don't know that it will, but I have no idea how to test it). I'll be optimistic and merge it, and we can revert if it breaks something.

dtest uses `/usr/bin/java` , and use `$JAVA_HOME/bin/java` as a fallback if `$JAVA_HOME` is defined.

this script is used by rpm and deb packaging. i just realized this usage. and tested using 
```console
./tools/toolchain/dbuild ninja dist-jmx-rpm
```

and then found the regression. sorry for the breakage.

PR #197, opened 2023-03-26
dist/redhat: support jre 11 instead of jre 8

tchaikov:
this change fixes the dependency required by scylla-jmx's rpm package.

it is part of the series preparing for jre 11. as we are observing
crashes in JVM provided by OpenJDK 8, see https://github.com/scylladb/scylla-jmx/pull/193. since OpenJDK 8
is aged and only security bugs are accepted now. let's move over to
OpenJDK 11 for a better supported Java platform. since we've already
changed jdk8 to jdk 11 in https://github.com/scylladb/scylla-jmx/commit/19097f2d73d0c42e60f35d3d1fa4d231cfd4809b.
and install-dependencies.sh is used both for preparing developer's
environment and for setting up the dbuild docker image, which is
used for running dtest also in our CI. we need to update scylla-jmx's
rpm accordingly, otherwise the rpm package won't install in the
container setup using the dbuild image once it is updated with the
latest scylla-jmx's install-dependencies.sh. please note, debian/control
already depends on openjdk-11-jre as an alternative of openjdk-8-jre,
so we are fine.

please note, jre-11-headless is provided by java-11-openjdk-headless.

### DIFF
--- a/dist/redhat/scylla-jmx.spec
+++ b/dist/redhat/scylla-jmx.spec
@@ -10,7 +10,7 @@ Source0:        %{reloc_pkg}
 
 BuildArch:      noarch
 BuildRequires:  systemd-units
-Requires:       %{product}-server jre-1.8.0-headless
+Requires:       %{product}-server jre-11-headless
 AutoReqProv:    no
 
 %description
--- a/dist/redhat/scylla-jmx.spec
+++ b/dist/redhat/scylla-jmx.spec
@@ -10,7 +10,8 @@ Source0:        %{reloc_pkg}
 
 BuildArch:      noarch
 BuildRequires:  systemd-units
-Requires:       %{product}-server jre-11-headless
+Requires:       %{product}-server
+Requires:       jre-11-headless
 AutoReqProv:    no
 
 %description


tchaikov:
@nyh hi, Nadav, i think we also need this change as a part of the plan in  https://github.com/scylladb/scylla-jmx/pull/193#discussion_r1148516582

tchaikov:
@avikivity thanks for queuing the change to bump up jmx, but i think we still need this one.

nyh (approved):

tchaikov:
https://github.com/scylladb/scylladb/pull/13356

tchaikov:
> @nyh It's better to ask for this before merging (the PR can then be closed as it will refer to the wrong repository).
> 

i am confused. how can we create a pull request bumping up a submodule which is supposed to include a change without getting that change landed in the submodule's repo? i think i don't have the write access to scylladb/scylla-jmx repo, and, in scylladb/scylladb repo, it references the scylla-jmx using `../scylla-jmx`, so, IIUC, we need to merge this change first, no?

nyh:
> @nyh It's better to ask for this before merging (the PR can then be closed as it will refer to the wrong repository).

The author should have done something like this, manually, to test what he did in the submodule and how it fits the whole project.

But I think there is a limit to how pedantic we need to be. If something wrong was committed to the scylla-jmx submodule, it's not a disaster - it only becomes a disaster when the submodule is upgrade in scylla.git - and this part will be checked by CI.

> Or you can open it yourself and avoid the extra back-and-forth latency.

I prefer not to. If there's a problem seen in the result, I don't want to fix it myself, I want the original submodule-patch author to worry about it.

avikivity:
So it's best to ask the author for a CI run. I want to avoid a window where the submodule is not synchronized. First, it creates conversations like this. Second, what if another patch to scylla-tools-java comes in? Things start to pile up.

Issue #199, opened 2023-04-11
metrics: sstables_per_read_histogram returns number of sstables reads


Deexie:
EstimatedHistogram, which is responsible for sstablesPerRead column
in SSTablesPerReadHistogram table histogram, returns values indicating 
the number of sstables reads instead of the number of sstables.

mykaul:
We don't have the 'backport' candidate here - unsure what the process is later on, since we do wish to have this fix in earlier releases as well (do we update the submodule?).
@DoronArazii - need to follow up once this is merged.

DoronArazii:
> @DoronArazii - need to follow up once this is merged.

It's in master --> https://github.com/scylladb/scylladb/commit/50ee4033a92e395e70c08e64d60667bb2681fd2b

denesb:
> > @DoronArazii - need to follow up once this is merged.
> 
> It's in master --> [scylladb/scylladb@50ee403](https://github.com/scylladb/scylladb/commit/50ee4033a92e395e70c08e64d60667bb2681fd2b)

It is actually scylladb/scylladb@f82287a9afbdaeb87138a0918f05f4c52e9b5966.

PR #200, opened 2023-04-11
metrics: EstimatedHistogram::getValues() returns bucketOffsets

Deexie:
EstimatedHistogram is expected to return number of sstables as its value, but it returns number of sstables reads.

EstimatedHistogram::getValues() returns bucketOffsets which contains the number of sstables.

Fixes: #199.


bhalevy:
LGTM.

Is there a test to verify the fix?

bhalevy:
https://github.com/scylladb/scylla-dtest/pull/3113, right?

bhalevy (approved):

PR #202, opened 2023-04-12
install.sh: select java executable at runtime

### LINKED ISSUE #194: ccm fails to start a node because connection to jmx cannot be established
Seen in CI:

https://jenkins.scylladb.com/job/releng/job/Scylla-CI/4555/testReport/junit/cdc_test/TestCdc/Sanity_Tests___test_change_field_type_with_cdc_Single_cluster_/

```
self = <cdc_test.TestCdc object at 0x7f9ab9f9d0d0>
request = <FixtureRequest for <Function test_change_field_type_with_cdc[Single_cluster]>>
cluster_config = ClusterConfig(size=[3], replication="{'class': 'SimpleStrategy', 'replication_factor': 3}")

    @pytest.mark.next_gating
    def test_change_field_type_with_cdc(self, request, cluster_config):
>       self.schema_change_template(request, "ALTER TABLE ks.cf ALTER b TYPE blob",
                                    cluster_size=cluster_config.size, replication=cluster_config.replication)

cdc_test.py:449: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
cdc_test.py:414: in schema_change_template
    self.populate_sequentially(n=cluster_size)
cdc_test.py:76: in populate_sequentially
    node.start(wait_for_binary_proto=True, wait_other_notice=wait_other_notice)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <ccmlib.scylla_node.ScyllaNode object at 0x7f9ae4466410>
join_ring = True, no_wait = False, verbose = False, update_pid = True
wait_other_notice = False, replace_token = None, replace_address = None
replace_node_host_id = None
jvm_args = ['--api-address', '127.0.40.2', '--collectd-hostname', '05767ff5161b.node2']
wait_for_binary_proto = True, profile_options = None, use_jna = False
quiet_start = False

    def start(self, join_ring=True, no_wait=False, verbose=False,
              update_pid=True, wait_other_notice=None, replace_token=None,
              replace_address=None, replace_node_host_id=None, jvm_args=None, wait_for_binary_proto=None,
              profile_options=None, use_jna=False, quiet_start=False):
        """
        Start the node. Options includes:
          - join_ring: if false, start the node with -Dcassandra.join_ring=False
          - no_wait: by default, this method returns when the node is started
            and listening to clients.
            If no_wait=True, the method returns sooner.
          - wait_other_notice: if True, this method returns only when all other
            live node of the cluster
            have marked this node UP.
          - replace_token: start the node with the -Dcassandra.replace_token
            option.
          - replace_node_host_id: start the node with the
            --replace-node-first-boot option to replace a given node
            identified by its host_id.
          - replace_address: start the node with the deprecated
            --replace-address option.
    
        Extra command line options may be passed using the
        SCYLLA_EXT_OPTS environment variable.
    
        Extra environment variables for running scylla can be passed using the
        SCYLLA_EXT_ENV environment variable.
        Those are represented in a single string comprised of one or more
        pairs of "var=value" separated by either space or semicolon (';')
        """
        if wait_for_binary_proto is None:
            wait_for_binary_proto = self.cluster.force_wait_for_cluster_start and not no_wait
        if wait_other_notice is None:
            wait_other_notice = self.cluster.force_wait_for_cluster_start and not no_wait
        if jvm_args is None:
            jvm_args = []
    
        scylla_cassandra_mapping = {'-Dcassandra.replace_address_first_boot':
                                    '--replace-address-first-boot'}
        # Replace args in the form
        # ['-Dcassandra.foo=bar'] to ['-Dcassandra.foo', 'bar']
        translated_args = []
        new_jvm_args = []
        for jvm_arg in jvm_args:
            if '=' in jvm_arg:
                split_option = jvm_arg.split("=")
                e_msg = ("Option %s not in the form '-Dcassandra.foo=bar'. "
                         "Please check your test" % jvm_arg)
                assert len(split_option) == 2, e_msg
                option, value = split_option
                # If we have information on how to translate the jvm option,
                # translate it
                if option in scylla_cassandra_mapping:
                    translated_args += [scylla_cassandra_mapping[option],
                                        value]
                # Otherwise, just pass it as is
                else:
                    new_jvm_args.append(jvm_arg)
            else:
                new_jvm_args.append(jvm_arg)
        jvm_args = new_jvm_args
    
        if self.is_running():
            raise NodeError("%s is already running" % self.name)
    
        if not self.is_docker():
            for itf in list(self.network_interfaces.values()):
                if itf is not None and replace_address is None:
                    try:
                        common.check_socket_available(itf)
                    except Exception as msg:
                        print("{}. Looking for offending processes...".format(msg))
                        for proc in psutil.process_iter():
                            if any(self.cluster.ipprefix in cmd for cmd in proc.cmdline()):
                                print("name={} pid={} cmdline={}".format(proc.name(), proc.pid, proc.cmdline()))
                        raise msg
    
        marks = []
        if wait_other_notice:
            marks = [(node, node.mark_log()) for node in
                     list(self.cluster.nodes.values()) if node.is_live()]
    
        self.mark = self.mark_log()
    
        launch_bin = common.join_bin(self.get_path(), BIN_DIR, 'scylla')
        options_file = os.path.join(self.get_path(), 'conf', 'scylla.yaml')
    
        # TODO: we do not support forcing specific settings
        # TODO: workaround for api-address as we do not load it
        # from config file scylla#59
        conf_file = os.path.join(self.get_conf_dir(), common.SCYLLA_CONF)
        with open(conf_file, 'r') as f:
            data = yaml.safe_load(f)
        jvm_args = jvm_args + ['--api-address', data['api_address']]
        jvm_args = jvm_args + ['--collectd-hostname',
                               '%s.%s' % (socket.gethostname(), self.name)]
    
        # Let's add jvm_args and the translated args
    
        args = [launch_bin, '--options-file', options_file, '--log-to-stdout', '1'] + jvm_args + translated_args
    
        # Lets search for default overrides in SCYLLA_EXT_OPTS
        scylla_ext_opts = os.getenv('SCYLLA_EXT_OPTS', "").split()
        opts_i = 0
        orig_args = list(args)
        while opts_i < len(scylla_ext_opts):
            if scylla_ext_opts[opts_i].startswith("--scylla-manager="):
               opts_i += 1
            elif scylla_ext_opts[opts_i].startswith('-'):
                o = scylla_ext_opts[opts_i]
                opts_i += 1
                if '=' in o:
                    opt = o.replace('=', ' ', 1).split()
                else:
                    opt = [ o ]
                    while opts_i < len(scylla_ext_opts) and not scylla_ext_opts[opts_i].startswith('-'):
                        opt.append(scylla_ext_opts[opts_i])
                        opts_i += 1
                if opt[0] not in orig_args:
                    args.extend(opt)
    
        if '--developer-mode' not in args:
            args += ['--developer-mode', 'true']
        if '--smp' not in args:
            # If --smp is not passed from cmdline, use default (--smp 1)
            args += ['--smp', str(self._smp)]
        elif self._smp_set_during_test:
            # If node.set_smp() is called during the test, ignore the --smp
            # passed from the cmdline.
            args[args.index('--smp') + 1] = str(self._smp)
        else:
            # Update self._smp based on command line parameter.
            # It may be used below, along with self._mem_mb_per_cpu, for calculating --memory
            self._smp = int(args[args.index('--smp') + 1])
        if '--memory' not in args:
            # If --memory is not passed from cmdline, use default (512M per cpu)
            args += ['--memory', '{}M'.format(self._mem_mb_per_cpu * self._smp)]
        elif self._mem_set_during_test:
            # If node.set_mem_mb_per_cpu() is called during the test, ignore the --memory
            # passed from the cmdline.
            args[args.index('--memory') + 1] = '{}M'.format(self._mem_mb_per_cpu * self._smp)
        self._memory = self.parse_size(args[args.index('--memory') + 1])
        if '--default-log-level' not in args:
            args += ['--default-log-level', self.__global_log_level]
        if self.scylla_mode() == 'debug' and '--blocked-reactor-notify-ms' not in args:
            args += ['--blocked-reactor-notify-ms', '5000']
        # TODO add support for classes_log_level
        if '--collectd' not in args:
            args += ['--collectd', '0']
        if '--cpuset' not in args:
            args += ['--overprovisioned']
        if '--prometheus-address' not in args:
            args += ['--prometheus-address', data['api_address']]
        if replace_node_host_id:
            assert replace_address is None, "replace_node_host_id and replace_address cannot be specified together"
            args += ['--replace-node-first-boot', replace_node_host_id]
        elif replace_address:
            args += ['--replace-address', replace_address]
        args += ['--unsafe-bypass-fsync', '1']
    
        current_node_version = self.node_install_dir_version() or self.cluster.version()
        current_node_is_enterprise = parse_version(current_node_version) > parse_version("2018.1")
    
        # The '--kernel-page-cache' was introduced by
        # https://github.com/scylladb/scylla/commit/8785dd62cb740522d80eb12f8272081f85be9b7e from 4.5 version
        # and 2022.1 Enterprise version
        kernel_page_cache_supported = not current_node_is_enterprise and parse_version(current_node_version) >= parse_version('4.5.dev')
        kernel_page_cache_supported |= current_node_is_enterprise and parse_version(current_node_version) >= parse_version('2022.1.dev')
        if kernel_page_cache_supported and '--kernel-page-cache' not in args:
            args += ['--kernel-page-cache', '1']
        commitlog_o_dsync_supported = (
            (not current_node_is_enterprise and parse_version(current_node_version) >= parse_version('3.2'))
            or (current_node_is_enterprise and parse_version(current_node_version) >= parse_version('2020.1'))
            )
        if commitlog_o_dsync_supported:
            args += ['--commitlog-use-o-dsync', '0']
    
        # The '--max-networking-io-control-blocks' was introduced by
        # https://github.com/scylladb/scylla/commit/2cfc517874e98c5780c1b1b4c38440a8123f86f6 from 4.6 version
        # and 2022.1 Enterprise version
        max_networking_io_control_blocks_supported = not current_node_is_enterprise and parse_version(current_node_version) >= parse_version('4.6')
        max_networking_io_control_blocks_supported |= current_node_is_enterprise and parse_version(current_node_version) >= parse_version('2022.1.dev')
        if max_networking_io_control_blocks_supported and '--max-networking-io-control-blocks' not in args:
            args += ['--max-networking-io-control-blocks', '1000']
    
        ext_env = {}
        scylla_ext_env = os.getenv('SCYLLA_EXT_ENV', "").strip()
        if scylla_ext_env:
            scylla_ext_env = re.split(r'[; ]', scylla_ext_env)
            for s in scylla_ext_env:
                try:
                    [k, v] = s.split('=', 1)
                except ValueError as e:
                    print("Bad SCYLLA_EXT_ENV variable: {}: {}", s, e)
                else:
                    ext_env[k] = v
    
        message = "Starting scylla: args={} wait_other_notice={} wait_for_binary_proto={}".format(args, wait_other_notice, wait_for_binary_proto)
        self.debug(message)
    
        scylla_process = self._start_scylla(args, marks, update_pid,
                                            wait_other_notice,
                                            wait_for_binary_proto,
                                            ext_env)
        self._start_jmx(data)
    
        ip_addr, _ = self.network_interfaces['storage']
        jmx_port = int(self.jmx_port)
        if not self._wait_java_up(ip_addr, jmx_port):
            e_msg = "Error starting node {}: unable to connect to scylla-jmx port {}:{}".format(
                     self.name, ip_addr, jmx_port)
>           raise NodeError(e_msg, scylla_process)
E           ccmlib.node.NodeError: Error starting node node2: unable to connect to scylla-jmx port 127.0.40.2:7199

../scylla/.local/lib/python3.11/site-packages/ccmlib/scylla_node.py:650: NodeError
```

tchaikov:
before this change, a symbolic link to java runtime is created in the install directory of scylla-jmx, and scylla-jmx uses it to launch scylla-jmx-1.0.jar. and we only check for the expected java version when `install.sh` is called without `--packaging` specified. but, another important use case of `install.sh` is for creating relocatable package. in that case, the `--packaging` parameter is specified. so we always end up using `/usr/bin/java` for launching jmx jar if jmx is installed from the relocatable package.

but we do intend to use OpenJDK-11 for running scylla-jmx if it is available, no matter how scylla-jmx is installed.

so, in this change, instead of relying on user pointing `/usr/bin/java` to the appropriate java runtime, we just choose it by ourselves. to accomplish this goal,
`scripts/select-java` is extracted out of `install.sh`. this script picks JRE and execv it, and fails if neither JRE-8 nor JRE-11 is found. to minimize the impact to other parts of this project, while preserve a meaningful name in the source tree, the name `select-java` is used, while it is installed using the old name of `symlinks/scylla-jmx`, even it is not a symblink anymore.

also, the error mesage is adjusted to place "openjdk-11" before "openjdk-8" as a hint that openjdk-11 is prefered. this error mesage is only printed out in non-packaging mode.

Fixes #194
Signed-off-by: Kefu Chai <kefu.chai@scylladb.com>


tchaikov:
tested by running following commands
```console
$ ./tools/toolchain/dbuild ./configure.py --mode dev --disable-dpdk
$ ./tools/toolchain/dbuild ninja
```

and i checked the generated rpm package manually by:
```console
$ rpm2cpio scylla-jmx-5.3.0~dev-0.20230413.8d46436a87a6.noarch.rpm | cpio -idmv
$ ./usr/lib/scylla/jmx/select-java --version
openjdk 11.0.18 2023-01-17
OpenJDK Runtime Environment (Red_Hat-11.0.18.0.10-1.fc38) (build 11.0.18+10)
OpenJDK 64-Bit Server VM (Red_Hat-11.0.18.0.10-1.fc38) (build 11.0.18+10, mixed mode, sharing)
```

install.sh
```diff
@@ -20,7 +20,7 @@
 # along with Scylla.  If not, see <http://www.gnu.org/licenses/>.
 #
 
-set -e
+set -ex
```
denesb, line 23: is this a debugging leftover?

denesb (approved):

denesb:
Since this is a non-trivial change, I didn't immediately queue it in scylla.git, instead I opened a PR to have CI check it. See https://github.com/scylladb/scylladb/pull/13492.

tchaikov:
> Since this is a non-trivial change, I didn't immediately queue it in scylla.git, instead I opened a PR to have CI check it. See [scylladb/scylladb#13492](https://github.com/scylladb/scylladb/pull/13492).

@denesb hi Botond, thanks for your review and help.

install.sh
```diff
@@ -165,3 +166,5 @@ if $nonroot; then
 elif ! $without_systemd && ! $packaging; then
     systemctl --system daemon-reload
 fi
+
+echo done
```
tchaikov, line 170: it was added for debugging.

denesb: I already merged this. You can send a follow-up if you think this should be removed.

PR #204, opened 2023-04-14
install.sh: do not create /usr/scylla/jmx in nonroot mode

tchaikov:
in 08a6993750bb75102241b29937f17dd3a8c50ac1,
`install -m755 -d "$rusr"/lib/scylla/jmx` was added in the nonroot mode. this was wrong. and should only be added for the root mode, as in nonroot mode, we only have the write access to `$rprefix`, and `$ruse` is only valid in the non "nonroot" mode, where the script installs the package with the root user privileged to the system.

so, in this change, this step is dropped, as it is just wrong.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -155,7 +155,6 @@ if ! $nonroot; then
 fi
 
 if $nonroot; then
-    install -m755 -d "$rusr"/lib/scylla/jmx
     sed -i -e "s#/var/lib/scylla#$rprefix#g" "$rsysconfdir"/scylla-jmx
     sed -i -e "s#/etc/scylla#$rprefix/etc/scylla#g" "$rsysconfdir"/scylla-jmx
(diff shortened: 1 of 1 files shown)


tchaikov:
tested with following steps
```console
$ ./tools/toolchain/dbuild ./configure.py --mode dev --disable-dpdk
$ ./tools/toolchain/dbuild ninja
$ tar xzvf ./build/dev/dist/tar/scylla-unified-5.3.0~dev-0.20230413.8d46436a87a6.x86_64.tar.gz -C /tmp/scylla-unified
$ cd /tmp/scylla-unified/scylla-5.3.0~dev
$ ./install.sh --prefix /tmp/scylla-install --nonroot --supervisor --without-systemd
```

the last step completed without issues. before this change it fails with following error
```
install: cannot create directory ‘/lib/scylla’: Permission denied
```

Issue #206, opened 2023-04-16
scylla-jmx.service: failed during artifact tests


Annamikhlin:
Some artifact tests are failed with:
```
09:01:19  Command: '/usr/bin/nodetool  status '
09:01:19  Exit code: 1
09:01:19  Stdout:
09:01:19  Stderr:
09:01:19  nodetool: Failed to connect to '127.0.0.1:7199' - ConnectException: 'Connection refused (Connection refused)'.
```
Related to 
```
2023-04-16T05:59:02.939 artifacts-rocky8-jenkins-db-node-bcd77d4c-0-1 !NOTICE | systemd[1]: scylla-jmx.service: Main process exited, code=exited, status=1/FAILURE
2023-04-16T05:59:02.939 artifacts-rocky8-jenkins-db-node-bcd77d4c-0-1 !WARNING | systemd[1]: scylla-jmx.service: Failed with result 'exit-code'.
```
links to failed jobs:
https://jenkins.scylladb.com/job/scylla-master/job/artifacts/job/artifacts-ubuntu2004-arm-test/370/
https://jenkins.scylladb.com/job/scylla-master/job/artifacts/job/artifacts-ubuntu2004-test/929/
https://jenkins.scylladb.com/job/scylla-master/job/artifacts/job/artifacts-rocky8-test/209/

Annamikhlin:
@fgelcer - please the relevant logs

/cc: @fruch

fruch:
the actual failure (from ubuntu2004):
```
Apr 16 07:31:45 artifacts-ubuntu2004-jenkins-db-node-7d39823b-0-1 scylla-jmx[20387]:  using Hotspot JIT.
Apr 16 07:31:45 artifacts-ubuntu2004-jenkins-db-node-7d39823b-0-1 scylla-jmx[20387]:  .
Apr 16 07:31:45 artifacts-ubuntu2004-jenkins-db-node-7d39823b-0-1 scylla-jmx[20387]:  The packages are built using the IcedTea build support and patches
Apr 16 07:31:45 artifacts-ubuntu2004-jenkins-db-node-7d39823b-0-1 scylla-jmx[20387]:  from the IcedTea project.
Apr 16 07:31:45 artifacts-ubuntu2004-jenkins-db-node-7d39823b-0-1 scylla-jmx[20387]: Homepage: http://openjdk.java.net/
[...]

PR #207, opened 2023-04-17
treewide: build jar which targets jdk-11

tchaikov:
since we've moved to OpenJDK-11, let's build the jar targeting JDK-11.

the `--add-exports` change was created after referencing
https://docs.oracle.com/en/java/javase/16/migrate/migrating-jdk-8-later-jdk-releases.html#GUID-2F61F3A9-0979-46A4-8B49-325BA0EE8B66
without this change, we'd have build failures like
```
[ERROR] /home/kefu/dev/scylladb/tools/jmx/src/main/java/com/scylladb/jmx/utils/APIBuilder.java:[48,19] package com.sun.jmx.interceptor is not visible
[ERROR]   (package com.sun.jmx.interceptor is declared in module java.management, which does not export it)
```

so that the created jar file won't be compatible with JDK-8, this
helps us to identify the places where the JDK-8 is still used.

by passing `-source` to the javac compiler allows us to use the
Java-11 features. despite that we don't have plan to use any of
them, it is still a nice to have.

Refs #206
Signed-off-by: Kefu Chai <kefu.chai@scylladb.com>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -17,8 +17,8 @@
     <name>Scylla JMX</name>
 
     <properties>
-        <maven.compiler.target>1.8</maven.compiler.target>
-        <maven.compiler.source>1.8</maven.compiler.source>
+        <maven.compiler.target>11</maven.compiler.target>
+        <maven.compiler.source>11</maven.compiler.source>
     </properties>	    
 
     <dependencies>
@@ -57,6 +57,16 @@
                         </configuration>
                     </execution>
                 </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <configuration>
+                    <compilerArgs combine.children="append">
+                        <arg>--add-exports=java.management/com.sun.jmx.mbeanserver=ALL-UNNAMED</arg>
+                        <arg>--add-exports=java.management/com.sun.jmx.interceptor=ALL-UNNAMED</arg>
+                    </compilerArgs>
+                </configuration>
             </plugin>
         </plugins>
     </build>
--- a/scylla-apiclient/dependency-reduced-pom.xml
+++ b/scylla-apiclient/dependency-reduced-pom.xml
@@ -17,8 +17,8 @@
     </dependency>
   </dependencies>
   <properties>
-    <maven.compiler.source>1.8</maven.compiler.source>
-    <maven.compiler.target>1.8</maven.compiler.target>
+    <maven.compiler.source>11</maven.compiler.source>
+    <maven.compiler.target>11</maven.compiler.target>
   </properties>
 </project>
 
--- a/scylla-apiclient/pom.xml
+++ b/scylla-apiclient/pom.xml
@@ -10,8 +10,8 @@
     <name>Scylla REST API client</name>
 
     <properties>
-        <maven.compiler.target>1.8</maven.compiler.target>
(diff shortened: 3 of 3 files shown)


tchaikov:
tested by running following commands
```console
$ cd tools/jmx
$ ./reloc/build_reloc.sh --version $(<../../build/SCYLLA-PRODUCT-FILE)-$(sed 's/-/~/' <../../build/SCYLLA-VERSION-FILE)-$(<../../build/SCYLLA-RELEASE-FILE) --nodeps
$ ./scripts/scylla-jmx
```

tchaikov:
@avelanarius  hi Piotr, could you please help review this change?

mykaul:
If we build with Java 11, can we run with Java 8? I'm not sure.
If not, need to fix https://github.com/scylladb/scylla-jmx/blob/770c1f9fb06475fcad2337283e12188b5008a299/scripts/select-java#L21 and probably others?

tchaikov:
@mykaul hi Yaniv, thank you for your inputs.

> If we build with Java 11, can we run with Java 8? 

to my best knowledge, no.

> I'm not sure. If not, need to fix
> 
> https://github.com/scylladb/scylla-jmx/blob/770c1f9fb06475fcad2337283e12188b5008a299/scripts/select-java#L21
> 
> and probably others?

and the places of package dependencies. but before searching and replacing. i'd like have some reviews from our Java experts.

mykaul:
I wonder if we can do all those changes without doing them in parallel on https://github.com/scylladb/scylla-tools-java ? (of course, if each were to run in its own container, we wouldn't have had this issue)

avikivity:
I believe you can tell Java to build for an older version. But if we change our requires= to use java 11 then we can run on 11 too.

tchaikov:
> I believe you can tell Java to build for an older version. But if we change our requires= to use java 11 then we can run on 11 too.

if we build for an older version, in this question, it's JRE-8. then, that's practically partially reverts this very patch, which 

1. targets JRE-11: `-target 11`
2. allow using newer language features introduced by JDK-11: `-source 11`

after a second thought, probably we will not benefit from either of these changes though, if 

- we don't want to to ditch the JRE-8 support, or 
- we don't plan to use any of Java-11 features.

but i think the first item is still a nice-to-have. as we were suffering from the JRE-8 for a while, would be nice to just to be explicitly stop supporting it.

tchaikov:
> I wonder if we can do all those changes without doing them in parallel on https://github.com/scylladb/scylla-tools-java ? (of course, if each were to run in its own container, we wouldn't have had this issue)

the PR only changes the generated `scylla-jmx-1.0.jar`. and since we should always use jdk-11 to execute it. so it should has nothing to do with the java tools hosted in scylla-tools-java .

mykaul:
> > I wonder if we can do all those changes without doing them in parallel on https://github.com/scylladb/scylla-tools-java ? (of course, if each were to run in its own container, we wouldn't have had this issue)
> 
> the PR only changes the generated `scylla-jmx-1.0.jar`. and since we should always use jdk-11 to execute it before and after this change. so it should has nothing to do with the java tools hosted in scylla-tools-java .

In practice, it does - if we still support and deploy using Java 8, because this is what we have in the platform and the tools weren't changed to use 11 only. It seems to be that it needs to go hand in hand.

regevran:
I would like to drop this PR because of the following reasons:
- We are actively developing the native `nodetool`.
- Only developers that works outside of `dbuild` need this PR, and there are not many of them if at all.

PR #208, opened 2023-04-18
select-java: avoid exec multiple paths 

### LINKED ISSUE #206: scylla-jmx.service: failed during artifact tests

tchaikov:
both `dpkg -L .. | grep '/java$'` and `rpm -ql .. | grep '/java$`
can return multiple matches. and before this change, we just exec
the returned string even if it could include multiple paths.

after this change, when selecting java for debian and ubuntu,
use the path of `/usr/lib/jvm/java-11-openjdk-${arch}/bin/java` instead.
when selecting java for fedora and centos, we just return the first
path of `/usr/lib/jvm/jre-${version}-openjdk/bin/java` which exists.

this should address the failure like

```
Error: Could not find or load main class .usr.lib.jvm.java-8-openjdk-arm64.bin.java
```
when launching scylla-jmx as a service on a ubuntu box where
openjdk-8-jre-headless is installed, because:

```console
$ dpkg -L openjdk-8-jre-headless | grep '/java$'
/usr/lib/jvm/java-8-openjdk-amd64/jre/bin/java
/usr/lib/jvm/java-8-openjdk-amd64/bin/java
```
so the second line would be considered as the command line argument
pass to the first line.

Fixes #206

### DIFF
--- a/scripts/select-java
+++ b/scripts/select-java
@@ -1,43 +1,55 @@
 #!/bin/bash
 
-has_java=false
+function select_java_debian() {
+    for version in "11" "8"; do
+        pkg_name=openjdk-${version}-jre-headless
+        if ! dpkg-query --status ${pkg_name}; then
+            continue
+        fi
+        java=$(dpkg -L ${pkg_name} | grep '/java$')
+        if [ -n "$java" ]; then
+            echo $java
+            return
+        fi
+    done
+}
+
+function select_java_fedora() {
+    for version in "11" "1.8.0"; do
+        pkg_name=java-${version}-openjdk-headless
+        if ! rpm --quiet -q ${pkg_name}; then
+            continue
+        fi
+        java=$(rpm -ql ${pkg_name} | grep '/java$')
+        if [ -n "$java" ]; then
+            echo $java
+            return
+        fi
+    done
+}
+
+function select_java_others() {
+    javaver=$(/usr/bin/java -version 2>&1|head -n1|cut -f 3 -d " ")
+    if [[ "$javaver" =~ ^\"1.8.0 || "$javaver" =~ ^\"11.0. ]]; then
+        echo /usr/bin/java
+    fi
+}
+
 . /etc/os-release
 case "$ID" in
     ubuntu|debian)
-        for version in "11" "8"; do
-            pkg_name=openjdk-${version}-jre-headless
-            if ! dpkg-query --status ${pkg_name}; then
-                continue
-            fi
-            java=$(dpkg -L ${pkg_name} | grep '/java$')
-            if [ -n "$java" ]; then
-                break
-            fi
-        done
+        java=$(select_java_debian)
         ;;
     fedora|centos)
-        for version in "11" "1.8.0"; do
-            pkg_name=java-${version}-openjdk-headless
-            if ! rpm --quiet -q ${pkg_name}; then
-                continue
-            fi
-            java=$(rpm -ql ${pkg_name} | grep '/java$')
-            if [ -n "$java" ]; then
-                break
-            fi
-        done
+        java=$(select_java_fedora)
         ;;
 esac
 if [ -n "$java" ]; then
-    has_java=true
+    : # good
 elif [ -x /usr/bin/java ]; then
-    javaver=$(/usr/bin/java -version 2>&1|head -n1|cut -f 3 -d " ")
-    if [[ "$javaver" =~ ^\"1.8.0 || "$javaver" =~ ^\"11.0. ]]; then
-        java=/usr/bin/java
-        has_java=true
-    fi
+    java=$(select_java_others)
 fi
-if ! $has_java; then
+if [ -z "$java" ]; then
     exit 1
 fi
 
--- a/scripts/select-java
+++ b/scripts/select-java
@@ -1,13 +1,13 @@
 #!/bin/bash
 
 function select_java_debian() {
+    local java
+    local arch
+    arch=$(dpkg --print-architecture)
+
     for version in "11" "8"; do
-        pkg_name=openjdk-${version}-jre-headless
-        if ! dpkg-query --status ${pkg_name}; then
-            continue
-        fi
-        java=$(dpkg -L ${pkg_name} | grep '/java$')
-        if [ -n "$java" ]; then
+        java=/usr/lib/jvm/java-${version}-openjdk-${arch}/bin/java
+        if [ -e $java ]; then
             echo $java
             return
         fi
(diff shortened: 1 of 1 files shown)


tchaikov:
@nyh and @denesb hi Nadav and Botond, could you help review this change?

scripts/select-java
```diff
@@ -1,43 +1,55 @@
...
+        pkg_name=java-${version}-openjdk-headless
+        if ! rpm --quiet -q ${pkg_name}; then
+            continue
+        fi
+        java=$(rpm -ql ${pkg_name} | grep '/java$')
+        if [ -n "$java" ]; then
+            echo $java
+            return
+        fi
+    done
```
avikivity, line 27: I think we can use [[ -e /usr/lib/jvm/jre-11/bin/java ]] and [[ -e /usr/lib/jvm/jre-8/bin/java ]], it appears a standard way. No need to mess with packages.

tchaikov: Cool. Didn’t know this! Will update the fedora side.

tchaikov:
@avikivity updated both the debian and fedora sides. i think the debian packages also follow some sort of the convention,  so use it instead.

nyh (changes requested):
I left one comment about a shell error I think needs fixing.

But in addition to that, I wonder if the entire approach of looking for java couldn't been simpler with much less Fedora and Debian assumptions:

For example in Fedora, we know that Java is in /usr/lib/jvm/java/java-11 and /usr/lib/jvm/java/java-1.8.0. These are symbolic links to very specific versions, architectures, but it saves you the need to look for these versions or architectures. It also saves you the need to look for any RPM names, worry about whether the user installed the "headless" version or not headless of the RPM, and so on.

You can see an example in test/cql-pytest/run-cassandra. I simply try to run "java" from the path, and then /usr/lib/jvm/jre-11/bin/java, /usr/lib/jvm/jre-1.8.0/bin/java, and check if any of those exist and have an acceptable version. I don't need to deal with rpms. (Maybe Debian puts things in different places, but this just means adding a couple more known paths to this list).

scripts/select-java
```diff
@@ -1,43 +1,55 @@
...
 elif [ -x /usr/bin/java ]; then
-    javaver=$(/usr/bin/java -version 2>&1|head -n1|cut -f 3 -d " ")
-    if [[ "$javaver" =~ ^\"1.8.0 || "$javaver" =~ ^\"11.0. ]]; then
-        java=/usr/bin/java
-        has_java=true
-    fi
+    java=$(select_java_others)
 fi
-if ! $has_java; then
+if -z "$java"; then
```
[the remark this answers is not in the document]
tchaikov: @nyh i just fixed 1 minute ago.

nyh:
Oh, good, I see now the second patch indeed does what I suggested.
You could have merged the two patches, instead moving the functions in one patch and then completely replacing them in the second patch :-)

scripts/select-java
```diff
@@ -1,27 +1,24 @@
...
+    arch=$(dpkg --print-architecture)
+
     for version in "11" "8"; do
-        pkg_name=openjdk-${version}-jre-headless
-        if ! dpkg-query --status ${pkg_name}; then
-            continue
-        fi
-        java=$(dpkg -L ${pkg_name} | grep '/java$')
-        if [ -n "$java" ]; then
+        java=/usr/lib/jvm/java-11-openjdk-${arch}/bin/java
```
nyh, line 9: Debian doesn't have links to the name without the arch as Fedora has?

scripts/select-java
```diff
@@ -1,43 +1,55 @@
...
+            echo $java
+            return
+        fi
+    done
+}
+
+function select_java_others() {
+    javaver=$(/usr/bin/java -version 2>&1|head -n1|cut -f 3 -d " ")
+    if [[ "$javaver" =~ ^\"1.8.0 || "$javaver" =~ ^\"11.0. ]]; then
+        echo /usr/bin/java
```
nyh, line 31: By the way, instead of /usr/bin/java, you could have just used "java" (without a path), and assume that the user has it somewhere in the path. If you want the full path, you can use `which java`. But in any case, I agree /usr/bin/java will be the common case.

nyh (approved):

scripts/select-java
tchaikov: Ahh, I didn’t intend to change this part. But yes, `java` would be better. We can fix it in a follow-up change.

tchaikov:
> Oh, good, I see now the second patch indeed does what I suggested.
> You could have merged the two patches, instead moving the functions in one patch and then completely replacing them in the second patch :-)

Okay. I wanted to minimize the efforts  reviewers need to identify the exact change. Will merge changes like this in future.

mykaul:
> But in addition to that, I wonder if the entire approach of looking for java couldn't been simpler with much less Fedora and Debian assumptions:

Or we could put it in a container with the right, pre-defined dependencies...

nyh:
> Or we could put it in a container with the right, pre-defined dependencies...

Yes, the infamous https://en.wikipedia.org/wiki/Fundamental_theorem_of_software_engineering ;-)

tchaikov:
@avikivity @nyh hi Avi and Nadav, could you please merge this change? as without it, all artifacts related tests on ubuntu fail on the master branch.

scripts/select-java
denesb, line 9: Shouldn't you replace `11` with `${version}` here? You are checking for java 11 in each iteration of the loop.

tchaikov: @denesb hi Botond, thanks! fixed and repushed.

nyh (approved):

denesb (approved):

PR #209, opened 2023-04-18
dist/debian: drop dependencies on jdk-8

tchaikov:
this change mirrors https://github.com/scylladb/scylla-jmx/commit/f89670dfd80ed04860e1098f0fc6ef114793f31b,
since the oldest supported debian and its derivative are debian 10 and
ubuntu 18.04, both of which ship openjdk-11-jre-headless. and
oracle provides oracle-java11-set-default for them. so it's safe
to drop the optional dependencies on openjdk-8 and orcale-java8.

this would ensure that the scylla-jmx deb package's access to
jdk-11.

Fixes https://github.com/scylladb/scylla-jmx/issues/194

### DIFF
--- a/dist/debian/control.template
+++ b/dist/debian/control.template
@@ -8,7 +8,7 @@ Rules-Requires-Root: no
 
 Package: %{product}-jmx
 Architecture: all
-Depends: ${shlibs:Depends}, ${misc:Depends}, openjdk-8-jre-headless | openjdk-8-jre | oracle-java8-set-default | adoptopenjdk-8-hotspot-jre | openjdk-11-jre-headless | openjdk-11-jre |oracle-java11-set-default , %{product}-server
+Depends: ${shlibs:Depends}, ${misc:Depends}, openjdk-11-jre-headless | openjdk-11-jre |oracle-java11-set-default , %{product}-server
 Description: Scylla JMX server binaries 
  Scylla is a highly scalable, eventually consistent, distributed,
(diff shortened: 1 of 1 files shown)


tchaikov:
@nyh and @denesb hi Nadav and Botond, could you help review this change?

avikivity:
Even centos 7 has java 11, so how about we align the dependencies everywhere?

tchaikov:
@avikivity we already updated rpm side in https://github.com/scylladb/scylla-jmx/commit/f89670dfd80ed04860e1098f0fc6ef114793f31b . This change is the missing puzzle.

avikivity:
Ok. But we have to merge the other change first, otherwise an upgrade to a version that only contains this patch will break.

tchaikov:
> Ok. But we have to merge the other change first, otherwise an upgrade to a version that only contains this patch will break.

are you referring #208 ?

avikivity:
yes, I see it's merged now.

avikivity:
The second patch was broken, I dropped it.

```
19:43:47  dpkg-buildpackage: error: syntax error in debian/control at line 12: duplicate field Depends found
```

tchaikov:
> The second patch was broken, I dropped it.
> 
> ```
> 19:43:47  dpkg-buildpackage: error: syntax error in debian/control at line 12: duplicate field Depends found
> ```

sorry. fixed and created as #210

PR #213, opened 2023-05-10
sstableinfo: change the type of generation to string

tchaikov:
we will allow user to use UUID for the generation (identifier) of a SSTable. So to be compatible to both "long" and an "UUID", let's use "String" for the type of generation. this change requires the change on scylladb side, which should expose the generation as a string also in its "/storage_service/sstable_info" RESTful API.

### DIFF
--- a/src/main/java/org/apache/cassandra/service/SSTableInfo.java
+++ b/src/main/java/org/apache/cassandra/service/SSTableInfo.java
@@ -27,7 +27,7 @@ public class SSTableInfo {
     @XmlJavaTypeAdapter(type = Date.class, value = DateXmlAdapter.class)
     private Date timestamp;
 
-    private long generation;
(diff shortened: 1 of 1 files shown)


tchaikov:
the corresponding scylladb change is posted at https://github.com/scylladb/scylladb/pull/13834

tchaikov:
@raphaelsc hi Raphael, could you please help review this change?

tchaikov:
@bhalevy hi Benny, could advise whom shall i ping for reviewing this change?

denesb:
> @denesb why did you close this PR?

I merged it.

bhalevy:
> > @denesb why did you close this PR?
> 
> I merged it.

Oops, sorry, it must be too early for me then, since I didn't see that :)

PR #218, opened 2023-09-27
install-dependencies.sh: do not install weak dependencies

mykaul:
Especially for Java, we really do not need the tens of packages and MBs it adds, just because Java apps can be built and use sound and graphics and whatnot.

(Identical PR was sent earlier to scylla-tools-java repo)


tchaikov (approved):
thank you! this explains why i was still installing a bunch of fonts this morning even after applying my change in scylladb/scylladb.

Issue #226, opened 2023-11-20
Some requests break when there is a table whose name has a colon


nyh:
Normally, CQL table names may only have alphanumeric characters or underscores. The DynamoDB API also adds dash and dot as allowed characters. But when Alternator creates a GSI or LSI, the materialized view name will also contain a ":" or "!" characters, respectively.

It turns out that scylla-jmx has a bug when a table contains the colon character: In `src/main/java/org/apache/cassandra/db/ColumnFamilyStore.java` there is a function `doCheck()` there which loops over all tables returned by the `/column_family` REST API. This function calls the` getName()` function and croaks when the table name as a colon. What this function does is:
```java
    private static ObjectName getName(String type, String keyspace, String name) throws MalformedObjectNameException {
        return new ObjectName(
                "org.apache.cassandra.db:type=" + type + ",keyspace=" + keyspace + ",columnfamily=" + name);
    }
```
This is wrong if the given `name` may contain a colon. As https://docs.oracle.com/javase/8/docs/api/javax/management/ObjectName.html explains:

>  Each value associated with a key is a string of characters that is either unquoted or quoted.
> An unquoted value is a possibly empty string of characters which may not contain any of the characters comma, equals, colon, or quote.

So we need to quote the column family name in the above string, according to the quoting rules explained in https://docs.oracle.com/javase/8/docs/api/javax/management/ObjectName.html if it may contain any of these characters. Currently, only the colon is actually possible.

This issue cause a bug for `nodetool` users: Whenever there was an Alternator GSI somewhere in the database, `nodetool info` and `nodetool tablestats` stopped working, because both enumerate all the tables and come across this bug in their JMX request.

nyh:
Unfortunately, this opened a Pandora's box :-(

The "obvious" change:

```patch
     private static ObjectName getName(String type, String keyspace, String name) throws MalformedObjectNameException {
+        // The quote() call below is necessary for the table name because it might contain
+        // a colon (see https://github.com/scylladb/scylla-jmx/issues/226), but better be
+        // safe than sorry, so we use it on all values in the ObjectName string below,
+        // as required by the ObjectName documentation. We don't bother to quote "type"
+        // because we know it came from our code.
         return new ObjectName(
-                "org.apache.cassandra.db:type=" + type + ",keyspace=" + keyspace + ",columnfamily=" + name);
[...]

PR #227, opened 2023-11-20
Fix handling of table names with colon characters

nyh:
As noted in #226, nodetool breaks if a table name contains the colon character. This normally can't happen - CQL table names can only contain alphanumerics and underscores, and the DynamoDB API adds also dashes and dots - but neither allows a colon. But unfortunately, when Alternator has a secondary index (GSI or LSI), the name of the view _does_ get a colon (it looks like "tablename:viewname"). And this confuses JMX, which addresses objects using MBean names (see https://docs.oracle.com/javase/8/docs/api/javax/management/ObjectName.html), where a name that contains a colon needs to be quoted.

This small series fixes two things: First, some internal list of ObjectName objects needs to quote the names it holds to allow the colon character. Second, we should allow the external client - namely, nodetool - to send us a quoted name.

To complete this series I already have patches to two other repositories which I'll send shortly:
1. A patch to scylla-tools-java (i.e., nodetool) to use quoting when sending a table name with a colon to the JMX server.
2. A dtest patch to verify that before these patches, neither "nodetool info" or "nodetool tablestats" work, and after these patches, both work.

### DIFF
--- a/src/main/java/com/scylladb/jmx/utils/APIBuilder.java
+++ b/src/main/java/com/scylladb/jmx/utils/APIBuilder.java
@@ -425,6 +425,16 @@ public class APIBuilder extends MBeanServerBuilder {
             return (a == null) ? b == null : a.equals(b);
         }
 
+        // Unquote result of ObjectName.getKeyProperty(), if it was quoted
+        // by ObjectName.quote().
+        private static String maybe_unquote(String s) {
+            if (s == null || s.charAt(0) != '"') {
+                return s;
+            } else {
+                return ObjectName.unquote(s);
+            }
+        }
+
         @Override
         public boolean equals(Object o) {
             if (this == o) return true;
@@ -433,7 +443,7 @@ public class APIBuilder extends MBeanServerBuilder {
             }
             TableMetricParams oo = (TableMetricParams) o;
             return equal(name.getKeyProperty("keyspace"), oo.name.getKeyProperty("keyspace"))
-                    && equal(name.getKeyProperty("scope"), oo.name.getKeyProperty("scope"))
+                    && equal(maybe_unquote(name.getKeyProperty("scope")), maybe_unquote(oo.name.getKeyProperty("scope")))
                     && equal(name.getKeyProperty("name"), oo.name.getKeyProperty("name"))
                     && equal(name.getKeyProperty("type"), oo.name.getKeyProperty("type"));
         }
@@ -453,7 +463,7 @@ public class APIBuilder extends MBeanServerBuilder {
         @Override
         public int hashCode() {
             return safeAdd(hash(name.getKeyProperty("keyspace")),
-                    hash(name.getKeyProperty("scope")),
+                    hash(maybe_unquote(name.getKeyProperty("scope"))),
                     hash(name.getKeyProperty("name")),
                     hash(name.getKeyProperty("type")));
         }
--- a/src/main/java/org/apache/cassandra/db/ColumnFamilyStore.java
+++ b/src/main/java/org/apache/cassandra/db/ColumnFamilyStore.java
@@ -162,7 +162,10 @@ public class ColumnFamilyStore extends MetricsMBean implements ColumnFamilyStore
     }
 
     public ColumnFamilyStore(APIClient client, ObjectName name) {
-        this(client, name.getKeyProperty("type"), name.getKeyProperty("keyspace"), name.getKeyProperty("columnfamily"));
+        // The columnfamily values was quote()ed when creating the Objectname
+        // (see getName()), and name.getKeyProperty() returns it still quoted,
+        // so we need to unquote it back here.
+        this(client, name.getKeyProperty("type"), name.getKeyProperty("keyspace"), ObjectName.unquote(name.getKeyProperty("columnfamily")));
     }
 
     /** true if this CFS contains secondary index data */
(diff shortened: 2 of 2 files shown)


nyh:
I need help from the JMX or packaging gurus: How do we confirm that this patch doesn't break anything in the dtest suite? (it's supposed to fix a dtest which isn't yet in, but we need to confirm it doesn't break any existing test).

nyh:
See also pull requests in other repositories:
1. https://github.com/scylladb/scylla-tools-java/pull/355 - fix nodetool to be able to quote a table name with a colon in it (which the fix in the second patch in this PR can unquote).

elcallio (approved):

mykaul (approved):
LGTM. Was surprised that you need checks before 'unquote()' or it might throw an exception, but apparently that's the API.

nyh:
> LGTM. Was surprised that you need checks before 'unquote()' or it might throw an exception, but apparently that's the API.

Even after beeing knee-deep in this code for many hours, I'm not sure I fully understood what is really happening here or why it was designed this way. Why does unquote() even need to exist? After you used quote() to create a valid ObjectName, why doesn't the `getKeyProperty()` method retrieve the original unquoted string - why does the caller need to call unquote()? Anyway, experiments shows that this is what happens - if you used quote() to create the ObjectName, the quote characters are saved and returned by GetKeyProperty() :-(

All I can say is that I'm happy we are planning to get of this Java code eventually...

denesb (approved):

denesb:
Submodule update: 282dc0ae6d2c0486c65d35628b5ed0b7a94f7279

nyh:
> I need help from the JMX or packaging gurus: How do we confirm that this patch doesn't break anything in the dtest suite? (it's supposed to fix a dtest which isn't yet in, but we need to confirm it doesn't break any existing test).

@denesb did you do any such confirmation before merging this PR?
I see now next promotion failing on nodetool tests https://jenkins.scylladb.com/job/scylla-master/job/next/6846/#showFailuresLink - might be related :-(

denesb:
No, I just pushed it to next.

nyh:
I can reproduce a failure of
```
PYTHONPATH=../scylla-ccm pytest --cassandra-dir=$HOME/scylla/build/dev nodetool_additional_test.py::TestNodetool::test_get_sstable
```
I'll try to debug it. The curious thing is that obviously my patch did not kill en-masse all the nodetool tests, it just broke two specific tests. I'll need to debug now why.

Do you want to revert the Scylla patch that updates the submodule and leave the JMX patch in and I'll send a followup patch? Or do you want to revert the scylla-jmx patch as well?

Also **I repeat my question for the JMX and CI gurus** - don't we have CI runs for JMX patches? How do we prevent these cases from happening in the future?

denesb:
> Do you want to revert the Scylla patch that updates the submodule and leave the JMX patch in and I'll send a followup patch? Or do you want to revert the scylla-jmx patch as well?

The submodule update is is only in next, it can simply be dequeued.

> Also I repeat my question for the JMX and CI gurus - don't we have CI runs for JMX patches? How do we prevent these cases from happening in the future?

There is no CI for tools/* submodules, we merge to master directly. Also, there is usually no CI for submodule updates, unless the maintainer thinks the submodule update is risky, and opens a PR for CI. As you can see, the maintainer will sometimes misjudge the situation.

mykaul:
> Also **I repeat my question for the JMX and CI gurus** - don't we have CI runs for JMX patches? How do we prevent these cases from happening in the future?

I suspect we'll get rid of JMX sooner than we'll have tests for it :(

nyh:
Sent a followup: https://github.com/scylladb/scylla-jmx/pull/230

fruch:
> I can reproduce a failure of
> ```
> PYTHONPATH=../scylla-ccm pytest --cassandra-dir=$HOME/scylla/build/dev nodetool_additional_test.py::TestNodetool::test_get_sstable
> ```
> I'll try to debug it. The curious thing is that obviously my patch did not kill en-masse all the nodetool tests, it just broke two specific tests. I'll need to debug now why.
> 
> Do you want to revert the Scylla patch that updates the submodule and leave the JMX patch in and I'll send a followup patch? Or do you want to revert the scylla-jmx patch as well?
> 
> Also **I repeat my question for the JMX and CI gurus** - don't we have CI runs for JMX patches? How do we prevent these cases from happening in the future?

There is a way to run all of dtest tests, but you first build using the byo a unified package and then give to a dtest job to run all tests.

Other option is to run just the gating do a PR to Scylla with submodule change and let it run, I do it multiple time for cqlsh submodule, and it's mostly enough

nyh:
> There is a way to run all of dtest tests, but you first build using the byo a unified package and then give to a dtest job to run all tests.

My problem is that because I don't do those things often, I have no idea how to do what you just said, and no idea where there is documentation now how to do it.
 
> Other option is to run just the gating do a PR to Scylla with submodule change and let it run, I do it multiple time for cqlsh submodule, and it's mostly enough

Yes, this is a good thing to do before merging the submodule change. But in some sense it's a bit too late - it means we already merged the broken patch into scylla-jmx (for example), and yes, we won't have Scylla use the broken patch but it's still in the repository. But on second thought - maybe it's fine and we don't need to improve (and complicate) the process. We don't need to be fanatics on perfect code and bisectability on every small repository - maybe it's not a disaster that scylla-jmx got a buggy patch, and then a day later it got a second patch to fix it. Just as long as the buggy version isn't actually used in the Scylla parent module (and it isn't).

fruch:
> > There is a way to run all of dtest tests, but you first build using the byo a unified package and then give to a dtest job to run all tests.
> 
> My problem is that because I don't do those things often, I have no idea how to do what you just said, and no idea where there is documentation now how to do it.
> 
> > Other option is to run just the gating do a PR to Scylla with submodule change and let it run, I do it multiple time for cqlsh submodule, and it's mostly enough
> 
> Yes, this is a good thing to do before merging the submodule change. But in some sense it's a bit too late - it means we already merged the broken patch into scylla-jmx (for example), and yes, we won't have Scylla use the broken patch but it's still in the repository. But on second thought - maybe it's fine and we don't need to improve (and complicate) the process. We don't need to be fanatics on perfect code and bisectability on every small repository - maybe it's not a disaster that scylla-jmx got a buggy patch, and then a day later it got a second patch to fix it. Just as long as the buggy version isn't actually used in the Scylla parent module (and it isn't).

In some case I'm opening a PR pointing to my fork of the submodule, before merging, and opening a PR just for testing, and point it back to the actual repo/branch, when I confirmed it's working.

again maybe a bit convoluted and long, and can confuse the maintainers, but it get me a CI run ontop of my changes, with the least amount of effort. (but with not much control, since I can't change dtest/ccm in such run, or the exact test that would run)

denesb:
Submodule update PR: https://github.com/scylladb/scylladb/pull/16146

PR #229, opened 2023-11-20
StorageService: Normalize endpoint inetaddress strings to java form

### LINKED ISSUE #228: Endpoint strings in StorageService does not follow Java formatting for IPv6 nodes
This causes nodetool to break when mixing API:s returning actual IP addresses with info mapped by string.
Example is `nodetool status`.

Refs https://github.com/scylladb/scylladb/issues/16039

While this is really broken nodetool code, the problem is probably easier to fix here, in Scylla-JMX, as a stopgap until we can fully replace nodetool.

elcallio:
Fixes #228

Endpoint strings in StorageService does not follow Java formatting for IPv6 nodes. This causes nodetool to break when mixing API:s returning actual IP addresses with info mapped by string. Example is nodetool status.

Refs scylladb/scylladb#16039

While this is really broken nodetool code, the problem is probably easier to fix here, in Scylla-JMX, as a stopgap until we can fully replace nodetool.

### DIFF
--- a/src/main/java/org/apache/cassandra/service/StorageService.java
+++ b/src/main/java/org/apache/cassandra/service/StorageService.java
@@ -163,6 +163,40 @@ public class StorageService extends MetricsMBean implements StorageServiceMBean,
         logger.finest(str);
     }
 
+    private static String normalizeInetAddressString(String s) {
+        try {
+            return InetAddress.getByName(s).getHostAddress();
+        } catch (UnknownHostException e) {
+            return s;
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


avikivity:
@denesb please check native nodetool compatibility too

denesb:
> @denesb please check native nodetool compatibility too

`nodetool status` is not implemented yet.

avikivity:
ok, but please review this, since you're now our nodetool expert

denesb (approved):

denesb:
Submodule update PR: https://github.com/scylladb/scylladb/pull/16146

PR #232, opened 2024-01-05
dist/debian: do not use substvar of ${shlib:Depends}

tchaikov:
this subvar is exposed by `dpkg-shlibdeps`, which checks executables for their shared library dependencies, and expose them as `shlib:Depends` substitution variable.

but in the case of jmx, in place of executables, we package scylla-jmx-1.0.jar, which is a java jar archive. so we don't need to add dependencies for it.

also, we don't run `dpkg-shlibdeps` at all. that's why dpkg-gencontrol complains like:

```
dpkg-gencontrol: warning: Depends field of package scylla-jmx:
substitution variable ${shlibs:Depends} used, but is not defined
```

so, let's just drop this subst var.

### DIFF
--- a/dist/debian/control.template
+++ b/dist/debian/control.template
@@ -8,7 +8,7 @@ Rules-Requires-Root: no
 
 Package: %{product}-jmx
 Architecture: all
-Depends: ${shlibs:Depends}, ${misc:Depends},
+Depends: ${misc:Depends},
          openjdk-11-jre-headless | openjdk-11-jre | oracle-java11-set-default,
          %{product}-server
 Description: Scylla JMX server binaries 


tchaikov:
@syuu1228 hi Takuya, could you please help review this change? it should help to silence the warning like
```
11:19:35  dpkg-gencontrol: warning: Depends field of package scylla-jmx: substitution variable ${shlibs:Depends} used, but is not defined
```

see https://jenkins.scylladb.com/job/scylla-master/job/scylla-ci/5564/console

tchaikov:
@syuu1228 ping?

tchaikov:
@syuu1228 ping?

tchaikov:
@syuu1228 ping?

tchaikov:
@syuu1228 ping?

denesb:
@tchaikov this repo has its hours counted: it is on the way out. I don't think there is any point trying to improve it further.

tchaikov:
@denesb hi Botond, thank you for your insights. partially agreed. because it's still used by old branches. and it takes human cycles to parse the warnings -- i check every single warning in the build logging messages actually when the built fails. so it is still relevant for the supported branches, i think.

denesb:
Thanks for the explanation @tchaikov, good point.

syuu1228 (approved):
Seems like unneeded dependency for scylla-jmx, LGTM

PR #238, opened 2024-06-10
dist: support building packages in Github Actions

syuu1228:
Support building packages in Github Actions.
This includes to add support building in Ubuntu, since Github Actions only provides Ubuntu based image for Linux build.

Note that we don't use install-dependencies.sh on Github Actions, since we will use preinstalled Java to speed up installing dependencies.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,59 @@
+on:
+  push:
+    branches:
+      - master
+      - branch-*
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    name: Build packages
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: '11'
+          cache: 'maven'
+          cache-dependency-path: 'scylla-jmx-parent/pom.xml'
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y devscripts debhelper
+      - name: Build relocatable package
+        run: |
+          ./reloc/build_reloc.sh --nodeps
+          echo "VERSION=$(./SCYLLA-VERSION-GEN)" >> "$GITHUB_ENV"
+      - name: Build RPM
+        run: |
+          ./reloc/build_rpm.sh --reloc-pkg build/scylla-jmx-${VERSION}.noarch.tar.gz
+      - name: Build DEB
+        run: |
+          ./reloc/build_deb.sh --reloc-pkg build/scylla-jmx-${VERSION}.noarch.tar.gz
+      - name: Upload relocatable package
+        uses: actions/upload-artifact@v4
+        with:
+          name: scylla-jmx-relocatable-${{env.VERSION}}
+          path: build/scylla-jmx-*.noarch.tar.gz
+          if-no-files-found: error
+      - name: Upload RPM
+        uses: actions/upload-artifact@v4
+        with:
+          name: scylla-jmx-rpm-${{env.VERSION}}
+          path: build/redhat/RPMS/noarch/scylla-jmx-*.noarch.rpm
+          if-no-files-found: error
+      - name: Upload DEB
+        uses: actions/upload-artifact@v4
+        with:
+          name: scylla-jmx-deb-${{env.VERSION}}
+          path: build/debian/scylla-jmx_*_all.deb
+          if-no-files-found: error
+      - name: Publish packages to release
+        uses: softprops/action-gh-release@v2
+        if: startsWith(github.ref, 'refs/tags/')
+        with:
+          generate_release_notes: true
+          files: |
+            build/scylla-jmx-*.noarch.tar.gz
+            build/redhat/RPMS/noarch/scylla-jmx-*.noarch.rpm
+            build/debian/scylla-jmx_*_all.deb
--- a/SCYLLA-VERSION-GEN
+++ b/SCYLLA-VERSION-GEN
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 PRODUCT=scylla
 VERSION=666.development
--- a/dist/redhat/scylla-jmx.spec
+++ b/dist/redhat/scylla-jmx.spec
@@ -9,7 +9,6 @@ URL:            http://www.scylladb.com/
 Source0:        %{reloc_pkg}
 
 BuildArch:      noarch
-BuildRequires:  systemd-units
 Requires:       %{product}-server
 Requires:       jre-11-headless
 AutoReqProv:    no
--- a/install-dependencies.sh
+++ b/install-dependencies.sh
@@ -20,7 +20,8 @@
 . /etc/os-release
 
 if [ "$ID" = "ubuntu" ] || [ "$ID" = "debian" ]; then
-    apt -y install maven openjdk-11-jdk-headless
+    apt-get update
+    apt-get install -y maven openjdk-11-jdk-headless git rpm devscripts debhelper fakeroot dpkg-dev
 elif [ "$ID" = "fedora" ] || [ "$ID" = "centos" ]; then
-    dnf install -y maven java-11-openjdk-headless
+    dnf install -y maven java-11-openjdk-headless git rpm-build devscripts debhelper fakeroot dpkg-dev
 fi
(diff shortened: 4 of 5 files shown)


syuu1228:
And here's how it looks like on Actions page:
https://github.com/syuu1228/scylla-jmx/actions/runs/9449470120

.github/workflows/build.yml
```diff
@@ -0,0 +1,50 @@
+on:
+  push:
+    branches:
```
yaronkaikov, line 3: we probably need to run this workflow for every push to `master` and `branch-`

so you can add something like
```
branches:
  - master
  - branch-*.*
 ```

File: SCYLLA-VERSION-GEN [hunk 2]
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 PRODUCT=scylla
 VERSION=666.development
```
yaronkaikov, line 4: What about `VERSION` ? i guess we should start updating this value to match Scylla repo, right?

yaronkaikov (commented):
We should also publish the release when we create a new tag, we should use maybe https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository?tool=cli ,

yaronkaikov:
@avikivity Should we release `Scylla-JMX` every time we release Scylladb? or should it be completely independent? Maybe we should just create a release for every push to `branch-` ?

syuu1228:
> We should also publish the release when we create a new tag, we should use maybe https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository?tool=cli ,

Implemented.
It will be looks like:
https://github.com/syuu1228/scylla-jmx/releases/tag/scylla-5.0.2

Note that this will provide .rpm/.deb package but not yum/apt repositories.
If we want to provide repositories we cannot use Github release for the purpose.
What we need to do is basically same as we do in Jenkins today, build repositories, upload it to some HTTP server like Github Pages or Amazon S3, and point repository URLs on the document.

Or, alternatively we may able to host repositories in Fedora Copr / Launchpad PPA, push source package from Action to Fedora Copr / Launchpad PPA.

.github/workflows/build.yml
syuu1228: Added branch list.

SCYLLA-VERSION-GEN
syuu1228: `VERSION` is for define package version.

We use 666.development for master/next branch, and it need to be update to next release version when we create abranch or a tag, something like:
https://github.com/scylladb/scylla-jmx/commit/77d6f5e3e7d5e9be83df86e35c695c7bd8d8748e

When we specify VERSION to 5.0.2, build_{reloc,rpm,deb}.sh will automatically versioned like
scylla-jmx-5.0.2-20240611.d0cd907 ($VERSION-$DATE.$HASH).

> i guess we should start updating this value to match Scylla repo, right?

I think we do not need to match version with Scylla repo, it was required since our metapackage requires exact version number, like:
```
Requires:       %{product}-jmx = %{version}-%{release}
```
https://github.com/scylladb/scylladb/blob/branch-5.4/dist/redhat/scylla.spec#L14C1-L15C1

It was needed to install specific version of Scylla using the metapackage.
But now we dropped scylla-jmx dependency, I think now we can version scylla-jmx independently.

Note that even we can versioning independently, we need to keep increment from existing version. 
And latest available version is probably "scylla-jmx-6.0.0-0.20240606.a77615adf324":
https://downloads.scylladb.com/downloads/scylla/rpm/centos/scylladb-6.0/noarch/

So we can start from 6.0.0-0.<more recent date> (or 6.0.1 to make versioning cleaer).

avikivity:
> @avikivity Should we release `Scylla-JMX` every time we release Scylladb? or should it be completely independent? Maybe we should just create a release for every push to `branch-` ?

No. It should have its own independent release cycle. Realistically I expect one release and then nothing.

yaronkaikov:
> > We should also publish the release when we create a new tag, we should use maybe https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository?tool=cli ,
> 
> Implemented. It will be looks like: https://github.com/syuu1228/scylla-jmx/releases/tag/scylla-5.0.2
> 
> Note that this will provide .rpm/.deb package but not yum/apt repositories. If we want to provide repositories we cannot use Github release for the purpose. What we need to do is basically same as we do in Jenkins today, build repositories, upload it to some HTTP server like Github Pages or Amazon S3, and point repository URLs on the document.
> 
> Or, alternatively we may able to host repositories in Fedora Copr / Launchpad PPA, push source package from Action to Fedora Copr / Launchpad PPA.

I think it's fine, we should document it so that if someone needs the latest release for Scylla-JMX it will use the GitHub repo to download the relevant package

.github/workflows/build.yml
```diff
@@ -0,0 +1,63 @@
...
+      - master
+      - branch-*
+  pull_request:
```
yaronkaikov, line 6: Why do we need to trigger on `pull_request`? once we merge one it will go under the role above

syuu1228: Dropped the lines

avikivity:
> > We should also publish the release when we create a new tag, we should use maybe https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository?tool=cli ,
> 
> Implemented. It will be looks like: https://github.com/syuu1228/scylla-jmx/releases/tag/scylla-5.0.2
> 
> Note that this will provide .rpm/.deb package but not yum/apt repositories. If we want to provide repositories we cannot use Github release for the purpose. What we need to do is basically same as we do in Jenkins today, build repositories, upload it to some HTTP server like Github Pages or Amazon S3, and point repository URLs on the document.
> 
> Or, alternatively we may able to host repositories in Fedora Copr / Launchpad PPA, push source package from Action to Fedora Copr / Launchpad PPA.

Best would be to integrate it into the regular yum/apt repositories that contain scylla.rpm. Perhaps a separate task.

yaronkaikov:
> > We should also publish the release when we create a new tag, we should use maybe https://docs.github.com/en/repositories/releasing-projects-on-github/managing-releases-in-a-repository?tool=cli ,
> 
> Implemented. It will be looks like: https://github.com/syuu1228/scylla-jmx/releases/tag/scylla-5.0.2
> 
> Note that this will provide .rpm/.deb package but not yum/apt repositories. If we want to provide repositories we cannot use Github release for the purpose. What we need to do is basically same as we do in Jenkins today, build repositories, upload it to some HTTP server like Github Pages or Amazon S3, and point repository URLs on the document.
> 
> Or, alternatively we may able to host repositories in Fedora Copr / Launchpad PPA, push source package from Action to Fedora Copr / Launchpad PPA.

@syuu1228 The release should be only when we push tags not on every push, how is it implemented?

syuu1228:
> @syuu1228 The release should be only when we push tags not on every push, how is it implemented?

It is implemented as if condition on "Publish packages to release":
```if: startsWith(github.ref, 'refs/tags/')```
https://github.com/scylladb/scylla-jmx/blob/48b097129426840fdff9705a499f5008ae841281/.github/workflows/build.yml#L53C8-L53C49

**Comment by syuu1228** [comment]
> Best would be to integrate it into the regular yum/apt repositories that contain scylla.rpm. Perhaps a separate task.

I guess we can append jmx repository on existing scylla .repo / .list something like this:
```
[scylla]
name=Scylla for Enterprise Linux $releasever - $basearch
baseurl=<URL>
enabled=1
gpgcheck=1
gpgkey=<GPG>

[scylla-generic]
name=Scylla for centos $releasever
baseurl=<URL>
enabled=1
gpgcheck=1
gpgkey=<GPG>

[scylla-jmx]
name=Scylla JMX
baseurl=<URL for JMX>
enabled=1
gpgcheck=1
gpgkey=<GPG for JMX>
```

But to do so we need construct yum/apt repository somewhere else as I commented above:
https://github.com/scylladb/scylla-jmx/pull/238#issuecomment-2161824249
And I agree it should be separated task.

yaronkaikov:
@avikivity Can we please merge this?

denesb:
Submodule update: scylladb/scylladb@2556e902b1c8fd0638a316c4ab5fc25e28a7011f.

yaronkaikov:
Ammm @denesb Why should we update the submodule? we don't use it anymore, right?

yaronkaikov:
> Ammm @denesb Why should we update the submodule? we don't use it anymore, right?

or should we keep updating it until we remove it from Scylla?

denesb:
> > Ammm @denesb Why should we update the submodule? we don't use it anymore, right?
> 
> or should we keep updating it until we remove it from Scylla?

Yes, I think so.

PR #242, opened 2024-07-18
install.sh: fix incorrect permission on strict umask

syuu1228:
On scylla main repo, we had problem that install.sh install files with incorrect permission on strict umask:
 - scylladb/scylladb#10752
 - scylladb/scylladb#14383

To fix this, we need to "chmod" on newly created files.

Also, we need to run /usr/bin/install with "-Z" to avoid SELinux permission issue:
 - scylladb/scylladb#8589

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -80,6 +80,10 @@ while [ $# -gt 0 ]; do
     esac
 done
 
+install() {
+    command install -Z "$@"
+}
+
 check_usermode_support() {
     user=$(systemctl --help|grep -e '--user')
     [ -n "$user" ]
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -143,6 +148,7 @@ User=
 Group=
 WorkingDirectory=$rprefix
 EOS
+    chmod 644 "$rsystemd"/scylla-jmx.service.d/nonroot.conf
```
avikivity, line 151: This is slightly wrong. If we're not root, the user is installing into their own directory, and we don't need group permissions. If umask removes group permissions then they probably want it that way.

However, fixing this is way more trouble than it's worth. Probably no one installs without root these days, it's all docker or package install. So I'll merge it as is.

PR #243, opened 2024-09-05
version: update scylla-jmx version to: 6.2.0

syuu1228:
Now scylla-jmx is not part of scylla main repo, we need to have own version numbers and release tags.
Since we already released Scylla with scylla-jmx until 6.1.x, so we can start new version from 6.2.0.

### DIFF
--- a/SCYLLA-VERSION-GEN
+++ b/SCYLLA-VERSION-GEN
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 PRODUCT=scylla
-VERSION=666.development
+VERSION=6.2.0
 
 if test -f version
 then


syuu1228:
This is related with https://github.com/scylladb/scylladb/pull/17969#issuecomment-2329856684

yaronkaikov (approved):

syuu1228:
Note that not just increment version number, we also need add a tag something like: scylla-6.2.0
to releasing artifacts.

syuu1228:
@avikivity ping, as I wrote on the thread https://github.com/scylladb/scylladb/pull/17969#issuecomment-2337450818, this is needed for publishing release page for scylla-jmx.

denesb:
Merged. Created and pushed a tag as well: https://github.com/scylladb/scylla-jmx/releases/tag/scylla-6.2.0.

denesb:
I suppose there is no point in updating this submodule reference in scylladb.git, seeing that we are just about to drop said reference.

PR #246, opened 2024-10-01
dist: Invoke GitHub Actions correctly when new tag pushed

### LINKED ISSUE #245: GitHub Actions doesn't invoked when new tag pushed
Due to bug of the GitHub Actions script, it does not invoked when new tag pushed.
I realized that even scylla-6.2.0 tag is pushed, release page is not created as we expected.
I checked the script, and realized that we probably missing "tags:" on the trigger definition:
https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#running-your-workflow-only-when-a-push-of-specific-tags-occurs

syuu1228:
We are missing "tags:" on the trigger definition, need to add it.

Fixes #245

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -3,6 +3,8 @@ on:
     branches:
       - master
       - branch-*
+    tags:
+      - '*'
 
 jobs:
   build:


yaronkaikov (approved):

syuu1228:
Can we try pushing a tag again, to invoke GitHub Actions correctly?

PR #249, opened 2026-06-10
ci: pin GitHub Actions to commit SHAs

roydahan:
## Summary
- Pin all external GitHub Actions to full commit SHAs to reduce supply chain attack surface
- Upgrade outdated actions to their latest versions

**This PR was generated automatically. Please verify that GitHub Actions work as expected with these changes before merging.**

Reference: https://github.com/scylladb/scylladb/pull/29421

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,8 +11,8 @@ jobs:
     runs-on: ubuntu-latest
     name: Build packages
     steps:
-      - uses: actions/checkout@v4
-      - uses: actions/setup-java@v4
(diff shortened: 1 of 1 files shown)


yaronkaikov (approved):
LGTM , although i don't understand why to bother with this repo?  it will never get updated i think

denesb:
I agree (that there is not much point), but I still merged because it is just 1 minute.

mykaul:
> I agree (that there is not much point), but I still merged because it is just 1 minute.

Exactly the reason. We don't want a supply chain attack reproducing old versions that people may trust. This is what happened to Trivy.

yaronkaikov:
Unless you planning backport this it will not get into any release

